PR #15, opened 2021-02-10
Move TS types to dev dependencies

rzadp:


package.json
```diff
@@ -8,6 +8,8 @@
   "devDependencies": {
     "@types/chai": "^4.2.14",
     "@types/mocha": "^8.2.0",
+    "@types/react": "^17.0.0",
```
JustynaBroniszewska, line 11: I think we should keep them in dependencies because we will export some of react types? https://stackoverflow.com/questions/45176661/how-do-i-decide-whether-types-goes-into-dependencies-or-devdependencies

rzadp: I see, I did not realise this.

PR #21, opened 2021-02-10
🚢 Port useChainCalls

rzadp:
- Port the `useChainCalls` along with all required dependencies and the `chainState` provider
- Add a playground example usage of `useChainCalls` in the from of `useBlockMeta`
- Closes https://github.com/EthWorks/useDapp/issues/20


packages/core/src/providers/chainState/context.ts
```diff
@@ -0,0 +1,26 @@
+import { createContext, useContext } from 'react'
+import { ChainCall } from './callsReducer'
+import { ChainState } from './model'
+
+export const ChainStateContext = createContext<{
```
Marik-D, line 5: Might not be required in this PR, but we should combine this with BlockNumber context

Marik-D (approved):

packages/core/src/providers/chainState/context.ts
rzadp: I created https://github.com/EthWorks/useDapp/issues/23 to handle this separately.
There are two different block numbers, which is illustrated here:

https://github.com/EthWorks/dapp-boilerplate/blob/dapp-framework/packages/ui/src/components/page/BlockNumber.tsx#L11-L12

PR #22, opened 2021-02-10
❄️ Extract playground into a separate package

Marik-D:


packages/example/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@usedapp/example",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "start": "snowpack dev",
+    "build": "snowpack build",
+    "format": "prettier --write \"src/**/*.{js,jsx,ts,tsx}\"",
```
rzadp, line 8:
```suggestion
```

`prettier` with fixing is defined below.

rzadp (approved):

PR #36, opened 2021-02-16
Refactor useBlockMeta

Marik-D:


packages/core/src/hooks/useBlockMeta.ts
```diff
@@ -1,16 +1,12 @@
...
 
 export function useBlockMeta() {
   const address = useMulticallAddress()
-
-  const calls: ChainCall[] = [
-    { address, data: MultiCallABI.encodeFunctionData('getCurrentBlockTimestamp', []) },
-    { address, data: MultiCallABI.encodeFunctionData('getCurrentBlockDifficulty', []) },
-  ]
-  const [timestamp, difficulty] = useChainCalls(address ? calls : [])
+  const timestamp = useChainCall({ address, data: MultiCallABI.encodeFunctionData('getCurrentBlockTimestamp', []) })
```
sz-piotr, line 8: maybe this should actually be encoded into a constant?

packages/core/src/hooks/useBlockMeta.ts
```diff
@@ -1,16 +1,12 @@
...
+  const difficulty = useChainCall({ address, data: MultiCallABI.encodeFunctionData('getCurrentBlockDifficulty', []) })
```
sz-piotr, line 9: same here

sz-piotr (changes requested):

sz-piotr (approved):

PR #39, opened 2021-02-18
⏱️ Readonly provider

Marik-D:


packages/core/src/hooks/useChainCalls.ts
```diff
@@ -15,7 +15,7 @@ export function useChainCalls(calls: ChainCall[]) {
   })
 }
 
-export function useChainCall(call: ChainCall | false | undefined) {
+export function useChainCall(call: ChainCall | false | undefined | '') {
```
sz-piotr, line 18:
```suggestion
type Falsy = false | 0 | "" | null | undefined 
export function useChainCall(call: ChainCall | Falsy) {
```

packages/core/src/providers/Providers.tsx
```diff
@@ -29,3 +33,23 @@ function getLibrary(provider: any): Web3Provider {
   library.pollingInterval = 15000
   return library
 }
+
+const BACKGROUND_CHAIN_URLS = {
+  1: 'https://mainnet.infura.io/v3/62687d1a985d4508b2b7a24827551934',
```
marekkirejczyk, line 38: eee.... `id` in code?
Let's extract it to separate configuration file and use provider that by default not requires id.
Ethers have provider that work out of the box (EtherscanProvider I think).

Also:
Let's fill in urls for all test networks by default

packages/core/src/providers/Providers.tsx
```diff
@@ -18,6 +21,7 @@ export function Providers(props: Props) {
   return (
     <Web3ReactProvider getLibrary={getLibrary}>
       <BlockNumberProvider>
+        {props.backgroundChain && <BackgroundChainActivator chainId={props.backgroundChain} />}
```
marekkirejczyk, line 24: Why Background? ReadOnly maybe?
BackgroundChainActivator -> ReadOnlyActivator

packages/example/src/index.tsx
```diff
@@ -5,7 +5,7 @@ import { App } from './App'
 
 ReactDOM.render(
   <React.StrictMode>
-    <Providers>
+    <Providers backgroundChain={1}>
```
marekkirejczyk, line 8: `<Providers readOnlyChain={1}>`?

marekkirejczyk (changes requested):
See comments

marekkirejczyk (approved):

PR #41, opened 2021-02-21
📋 Introduce documentation, license, code of conduct etc

marekkirejczyk:


docs/source/_static/custom.css
```diff
@@ -0,0 +1,251 @@
+html {
```
rzadp, line 1: Is this ours or is it copied from somewhere?
If copied can we have a link or info where this is coming from?

LICENSE
```diff
@@ -0,0 +1,7 @@
+Copyright 2020 Ethworks sp z o.o.
```
rzadp, line 1:
```suggestion
MIT License

Copyright (c) 2020 Ethworks sp z o.o.
```

packages/core/README.md
```diff
@@ -50,9 +50,44 @@ function App() {
...
+[Install Sphinx](https://www.sphinx-doc.org/en/master/usage/installation.html) to build documentation:
+
+```sh
+cd docs
+make html
+```
+
+Before building documentation for the first time you may have to install required python packages:
+```sh
+pip install -r docs/requirements.txt
```
rzadp, line 88: Looks like Sphinx is not working with python 2.7. Should we say `pip3` to indicate that python3 is required?

```suggestion
pip3 install -r docs/requirements.txt
```

rzadp (approved):
Left a couple of suggestions.

docs/source/_static/custom.css
marekkirejczyk: Form Waffle :)

PR #48, opened 2021-02-26
📈 Add token list example

Marik-D:


packages/example/src/App.tsx
```diff
@@ -1,13 +1,38 @@
...
 
+  const [tokenList, setTokenList] = useState<any>()
+  useEffect(() => {
+    ;(async () => {
+      const res = await fetch('https://wispy-bird-88a7.uniswap.workers.dev/?url=http://erc20.cmc.eth.link')
+      setTokenList(await res.json())
+    })()
+  })
+
+  // TODO(marik-d): Fix this and use existing hook and currency value class.
```
JustynaBroniszewska, line 23: Yes, we could use `useTokenBalance`, but it is not ready yet. 😄

JustynaBroniszewska (approved):

PR #53, opened 2021-02-26
🥗 Add useEtherBalance

HubertNastaly:


packages/core/src/hooks/useEtherBalance.ts
```diff
@@ -0,0 +1,14 @@
+import { MultiCallABI } from '../constants'
+import { useChainCall } from './useChainCalls'
+import { useMulticallAddress } from './useMulticallAddress'
+import { BigNumber } from '@ethersproject/bignumber'
+
+export function useEtherBalance(address: string | undefined | null) {
```
rzadp, line 6: use` | Falsy`

packages/core/src/constants/chainId.ts
```diff
@@ -4,7 +4,7 @@ export enum ChainId {
   Rinkeby = 4,
   Goerli = 5,
   Kovan = 42,
-  xDai = 100,
+  xDai = 100
```
JustynaBroniszewska, line 7: Don't do that.. xd

rzadp (approved):

PR #69, opened 2021-02-28
📔  Add more documentation

marekkirejczyk:


docs/source/core.rst
```diff
@@ -66,51 +66,141 @@ useBlock
...
+
+Calls will be combined into a single multicall across all uses of *useChainCall* and *useChainCalls*.
+
+It is recommended to use `useContractCall`_ where applicable instead of this method.
+
+*Parameters*
+
+- ``call: ChainCall | Falsy`` - a single call, also see `ChainCall`_. A call can be `Falsy`, as it is important to keep the same ordering of hooks even if in given render cycle there might be not enough information to perfom a call.
+
+
```
JustynaBroniszewska, line 82:
```suggestion
Makes a call to a specific contract and returns the value. The hook will cause the component to refresh whenever a new block is mined and the value is changed.

Calls will be combined into a single multicall across all uses of *useChainCall* and *useChainCalls*.

It is recommended to use `useContractCall`_ where applicable instead of this method.

*Parameters*

- ``call: ChainCall | Falsy`` - a single call, also see `ChainCall`_. A call can be `Falsy`, as it is important to keep the same ordering of hooks even if in a given render cycle there might be not enough information to perform a call.


```

docs/source/core.rst
```diff
@@ -66,51 +66,141 @@ useBlock
...
 useChainCalls
 =============
 
+Makes multiple call to a specific contracts and returns values. Hook will cause component to refesh when values change.
```
JustynaBroniszewska, line 86:
```suggestion
Makes multiple calls to specific contracts and returns values. The hook will cause the component to refresh when values change.
```

docs/source/core.rst
```diff
@@ -66,51 +66,141 @@ useBlock
...
+
+Calls will be combined into a single multicall across all uses of *useChainCall* and *useChainCalls*.
+It is recommended to use `useContractCall`_ where applicable instead of this methods.
```
JustynaBroniszewska, line 89:
```suggestion
It is recommended to use `useContractCall`_ where applicable instead of this method.
```

docs/source/core.rst
```diff
@@ -66,51 +66,141 @@ useBlock
...
+
 *Parameters*
 
-- ``calls: ChainCall[]``
+- ``calls: ChainCall[]`` - list of calls, also see `ChainCall`_. Calls need to be in the same order accros component renders.
```
JustynaBroniszewska, line 93:
```suggestion
- ``calls: ChainCall[]`` - list of calls, also see `ChainCall`_. Calls need to be in the same order across component renders.
```

docs/source/core.rst
```diff
@@ -66,51 +66,141 @@ useBlock
...
+
+useContractCall
+===============
+Makes a call to a specific contract and returns value. Hook will cause component to refesh when new block is mined and return value changes.
+A syntax sugar for `useChainCall`_ that uses ABI, function name and arguments instead of raw data.
```
JustynaBroniszewska, line 98:
```suggestion
Makes a call to a specific contract and returns the value. The hook will cause the component to refresh when a new block is mined and the return value changes.
A syntax sugar for `useChainCall`_ that uses ABI, function name, and arguments instead of raw data.
```

docs/source/core.rst
```diff
@@ -66,51 +66,141 @@ useBlock
...
+
+**Parameters**
+
+- ``calls: ContractCall | Falsy`` - a single call to a contract , also see `ContractCall`_
+
+**Returns**
+
+- ``any[] | undefined`` - result of a call or undefined if call didn't return yet
```
JustynaBroniszewska, line 106:
```suggestion
- ``any[] | undefined`` - the result of a call or undefined if call didn't return yet
```

docs/source/core.rst
```diff
@@ -66,51 +66,141 @@ useBlock
...
+
+useContractCalls
+===============
+Makes calls to specific contracts and returns values. Hook will cause component to refesh when new block is mined and any of return values changes.
+A syntax sugar for `useChainCalls`_ that uses ABI, function name and arguments instead of raw data.
```
JustynaBroniszewska, line 111:
```suggestion
Makes calls to specific contracts and returns values. The hook will cause the component to refresh when a new block is mined and the return values change.
A syntax sugar for `useChainCalls`_ that uses ABI, function name, and arguments instead of raw data.
```

docs/source/core.rst
```diff
@@ -174,6 +291,32 @@ List of intended supported chains. If a user tries to connect to an unsupported
 **pollingInterval**
 Polling interval for a new block.
 
+ChainCall
+=========
+
+Represents a single call on the blokchain that can be included in multicall.
```
JustynaBroniszewska, line 297:
```suggestion
Represents a single call on the blockchain that can be included in multicall.
```

docs/source/getting-started.rst
```diff
@@ -59,10 +59,12 @@ Below is a simple example:
...
 
 
-Full example code is available `here <https://github.com/EthWorks/useDapp/tree/master/packages/example>`_.
+Example is available `here <https://usedapp-example.netlify.app/>`_ and full example code is available `here <https://github.com/EthWorks/useDapp/tree/master/packages/example>`_.
 
+Connecting to a network
+-----------------------
 
-First thing you need to do is set up **DAppPRovider** with optional config and wrap your whole App in it. You can read about config :ref:`here<config>`
+First thing you need to do is set up **DAppProvider** with optional config and wrap your whole App in it. You can read about config :ref:`here<config>`
```
JustynaBroniszewska, line 67:
```suggestion
The first thing you need to do is set up **DAppProvider** with optional config and wrap your whole App in it. You can read about config :ref:`here<config>`
```

docs/source/core.rst
JustynaBroniszewska: @marekkirejczyk

docs/source/core.rst
```diff
@@ -134,6 +224,34 @@ Returns connection state and functions that allow to manipulate the state.
...
+Returns a balance of a given token for a given address.
+
+**Parameters**
+
+- ``address: string | Falsy`` - address of an account
+- ``tokenAddress: string | Falsy`` - address of a token contract
+
+**Returns**
+
+- ``BigNumber | undefined`` - balance or undefined if address or token is *Falsy* or not connected
```
JustynaBroniszewska, line 241:
```suggestion
- ``balance: BigNumber | undefined`` - a balance which is BigNumber or undefined if address or token is *Falsy* or not connected
```

docs/source/core.rst
```diff
@@ -66,51 +66,141 @@ useBlock
...
+
+Returns ether balance of a given account.
+
+**Parameters**
+
+- ``address: string | Falsy`` - address of an account
+
+**Returns**
+
+- ``BigNumber`` - balance of the account or *undefined* if not connected to network or address is falsy value
```
JustynaBroniszewska, line 191:
```suggestion
- ``balance: BigNumber | undefined`` - a balance which is BigNumber of the account or *undefined* if not connected to network or address is a falsy value
```

JustynaBroniszewska (approved):

PR #81, opened 2021-03-22
💲 Create useTokenAllowance hook

nezouse:


packages/core/src/hooks/useTokenAllowance.ts
```diff
@@ -0,0 +1,23 @@
+import { BigNumber } from '@ethersproject/bignumber'
+import { ERC20Interface } from '../constants'
+import { Falsy } from '../model/types'
+import { useContractCall } from './useContractCall'
+
+export function useTokenAllowance(
+  ownerAddress: string | Falsy,
+  spenderAddress: string | Falsy,
+  tokenAddress: string | Falsy
```
marekkirejczyk, line 9: `tokenAddress` should be first argument (subject of call first, than parameters)

nezouse: I did it the same way as useTokenBalance, should we change both of them then?

marekkirejczyk (approved):

packages/core/src/hooks/useTokenAllowance.ts
marekkirejczyk: Strange. Let's change the order in another PR. I'll add a ticket for that.

PR #89, opened 2021-03-24
🚁 Add address helpers

### LINKED ISSUE #72: Address helpers
Address helpers: `addressEqual`, `addressCompare`, `shortenAddress`, `shortenIfAddress`

Implementation draft:
```ts
function addressEqual(address0: string, address1: string) {
  return utils.getAddress(address0) === utils.getAddress(address1)
}
```

```ts
function addressCompare(address0: string, address1: string) {
  const isGreater = BigNumber.from(address0).gte(BigNumber.from(address1))
  return isGreater ? 1 : -1
}
```

```ts
export function shortenAddress(address: string) {
  return address.substring(0, 6) + '...' + address.substring(address.length - 4)
}
```

nezouse:
Closes #72

### DIFF
--- a/packages/core/src/helpers/address.ts
+++ b/packages/core/src/helpers/address.ts
@@ -1,10 +1,13 @@
 import { utils } from 'ethers'
 import { BigNumber } from '@ethersproject/bignumber'
 
-export function shortenAddress(address: string) {
-  const formattedAddress = utils.getAddress(address)
-
-  return formattedAddress.substring(0, 6) + '...' + formattedAddress.substring(formattedAddress.length - 4)
+export function shortenAddress(address: string): string {
+  try {
+    const formattedAddress = utils.getAddress(address)
+    return formattedAddress.substring(0, 6) + '...' + formattedAddress.substring(formattedAddress.length - 4)
+  } catch {
+    throw new TypeError("Invalid input, address can't be parsed")
+  }
 }
 
 export function compareAddress(firstAddress: string, secondAddress: string) {
--- a/packages/core/src/helpers/address.ts
+++ b/packages/core/src/helpers/address.ts
@@ -11,3 +11,7 @@ export function compareAddress(firstAddress: string, secondAddress: string) {
   const isGreater = BigNumber.from(firstAddress).gte(BigNumber.from(secondAddress))
   return isGreater ? 1 : -1
 }
+
+export function addressEqual(firstAddress: string, secondAddress: string) {
+  return utils.getAddress(firstAddress) === utils.getAddress(secondAddress)
+}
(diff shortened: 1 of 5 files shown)


packages/core/src/helpers/address.ts
```diff
@@ -0,0 +1,17 @@
+import { utils } from 'ethers'
+import { BigNumber } from '@ethersproject/bignumber'
+
+export function shortenAddress(address: string) {
```
marekkirejczyk, line 4: Considerations:
1) Add validation (throw an error if not a valid address)
2) Add second helper (next PR) `shortenIfAddress(address: Falsy<string>)` which will shorten address or return empty string if falsy value

packages/core/test/helpers/address.test.ts
```diff
@@ -0,0 +1,83 @@
...
+      const address2 = '0xd1bbe67b8a28985f9a790bd2f13ebb42f0fc679c'
+
+      expect(addressEqual(address1, address2)).to.eq(false)
+    })
+
+    it('not an address', () => {
+      const address1 = 'im not an address'
+      const address2 = '0x2690a21899aaecea9fa832972677128ff8983dd8'
+
+      expect(() => addressEqual(address1, address2)).to.throw
```
marekkirejczyk, line 80: #checkexception

packages/core/test/helpers/address.test.ts
```diff
@@ -0,0 +1,83 @@
...
+      const address2 = '0x8ea9f79f41a20c6a7c724c1d808f9946b7ed620b'
+
+      expect(compareAddress(address1, address2)).to.eq(1)
+    })
+
+    it("can't parse address", () => {
+      const address1 = 'im not an address'
+      const address2 = '0xb293c3b2b4596824c57ad642ea2da4e146cca4cf'
+
+      expect(() => compareAddress(address1, address2)).to.throw
```
marekkirejczyk, line 57: #checkexception 
Always check what exception is thrown. 
Both in production and test code. 
Make sure it fails for the right reason, to make sure one (wrong) exception overshadows other (right) exception.
Also it is good documentation practice.

marekkirejczyk:
Missing docs

packages/core/test/helpers/address.test.ts
```diff
@@ -0,0 +1,83 @@
...
+      const address2 = '0xe24212440ad484f55997750cfae3e13ca1751283'
+
+      expect(compareAddress(address1, address2)).to.eq(-1)
+    })
+
+    it('both equal', () => {
+      const address1 = '0x8ea9f79f41a20c6a7c724c1d808f9946b7ed620b'
+      const address2 = '0x8ea9f79f41a20c6a7c724c1d808f9946b7ed620b'
+
+      expect(compareAddress(address1, address2)).to.eq(1)
```
marekkirejczyk, line 50: Should be zero.
See:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

marekkirejczyk (approved):

PR #95, opened 2021-03-26
Add hook for writing to blockchain

nezouse:
If API looks good I will add docs for it.
We probably need a better place for tests, but right now I just put it all in the same place.


packages/core/src/hooks/useContractFunction.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+  | {
+      status: 'Success'
+      chainId: ChainId
+      transaction: TransactionResponse
+      receipt: TransactionReceipt
+    }
+  | {
+      status: 'Fail'
+      transaction: TransactionResponse
```
marekkirejczyk, line 26: Should also contain `receipt`

packages/core/src/hooks/useContractFunction.ts
```diff
@@ -0,0 +1,86 @@
...
+  const send = async (...args: any[]) => {
+    if (!chainId) {
+      return
+    }
+
+    try {
+      transaction = await contractWithSigner[functionName](...args)
+      setState({ transaction, status: 'Mining', chainId })
+
+      const receipt = await transaction.wait()
```
marekkirejczyk, line 74: Hmm, I actually think that if transaction fails it will not throw an exception, but rather return receipt {..., success: false}.

Would you check?

packages/testing/examples/connectContractToSigner.test.ts
```diff
@@ -0,0 +1,45 @@
+import { ERC20Interface, connectContractToSigner, useEthers } from '@usedapp/core'
+import chai, { expect } from 'chai'
+import { MockProvider, solidity } from 'ethereum-waffle'
+import { Contract } from 'ethers'
+import { renderWeb3Hook } from '../src'
+
+chai.use(solidity)
+
+describe.only('useContractFunction', () => {
```
marekkirejczyk, line 9: `.only` 
Tip: better to use `yarn test -g 'useContractFunction'` , than `.only` prefix

packages/testing/examples/connectContractToSigner.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  })
+
+  it('takes signer from library', async () => {
+    const { result, waitForCurrent } = await renderWeb3Hook(() => useEthers(), { mockProvider })
+    await waitForCurrent((val) => val !== undefined)
+    const { library } = result.current
+
+    const connectedContract = connectContractToSigner(token, undefined, library)
+
+    expect(connectedContract.signer).to.be.not.null
```
marekkirejczyk, line 43: Why not 
`expect(connectedContract.signer).to.eq(library.getSigner())`

packages/core/src/hooks/useContractFunction.ts
marekkirejczyk: We made a wrong assumption based on ganache behaviour.

packages/testing/examples/useContractFunction.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+describe('useContractFunction', () => {
+  const mockProvider = new MockProvider()
+  const [deployer, spender] = mockProvider.getWallets()
+  let token: Contract
+
+  beforeEach(async () => {
+    token = await deployMockToken(deployer)
+  })
+
+  it('can approve erc20 token', async () => {
```
marekkirejczyk, line 20: Success

packages/testing/examples/useContractFunction.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+    })
+
+    await result.current.send(spender.address, 200)
+    await waitForCurrent((val) => val.state !== undefined)
+
+    expect(result.current.state.status).to.eq('Success')
+    expect(await token.allowance(deployer.address, spender.address)).to.eq(200)
+  })
+
+  it('it reverts with bad arguments', async () => {
```
marekkirejczyk, line 32: Exception

packages/testing/examples/useContractFunction.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+    await result.current.send()
+    await waitForCurrent((val) => val.state !== undefined)
+
+    expect(result.current.state.status).to.eq('Exception')
+    if (result.current.state.status === 'Exception') {
+      expect(result.current.state.errorMessage).to.eq('missing argument: passed to contract')
+    }
+  })
+
+  it('fails (revert)', async () => {
```
marekkirejczyk, line 46: `it('Fail (when transaction reverts)', ...`

packages/testing/examples/connectContractToSigner.test.ts
nezouse: It can probably be .to.deep.eq, but just to.eq won't work as it returns new object every time.

packages/core/src/hooks/useContractFunction.ts
nezouse: I will check it again, but I tested it on kovan and it was throwing on the example with not enough gas.

marekkirejczyk (approved):

PR #101, opened 2021-03-29
move ethereum-waffle to devDependencies

gasolin:
ethereum-waffle is the dev only package, should put in devDependencies


marekkirejczyk (approved):

marekkirejczyk:
Good catch :)

PR #103, opened 2021-03-30
📲 Add useTransactionsContext

nezouse:


packages/core/src/providers/transactions/model.ts
```diff
@@ -0,0 +1,18 @@
+import { TransactionReceipt, TransactionResponse } from '@ethersproject/providers'
+import { ChainId } from '../../constants'
+
+export type TransactionWithChainId = TransactionResponse & {
+  chainId: ChainId
+}
+
+export interface StoredTransaction {
```
marekkirejczyk, line 8: `TransactionItem` :D

marekkirejczyk (approved):

PR #107, opened 2021-03-31
🏎 Optimise useTransactions hook

nezouse:


packages/core/src/providers/transactions/provider.tsx
```diff
@@ -44,34 +44,40 @@ export function TransactionProvider({ children }: Props) {
...
       }
 
-      const chainTransactions = transactions[chainId] ?? []
-      const newTransactions = await Promise.all(
-        chainTransactions.map(async (tx) => {
-          if (tx.receipt) {
-            return tx
-          }
+      const checkTransaction = async (tx: StoredTransaction) => {
+        if (tx.receipt || !shouldCheck(blockNumber, tx)) {
```
marekkirejczyk, line 48: Duplication: `tx.receipt` with first line of `shouldCheck`

packages/core/src/providers/transactions/provider.tsx
```diff
@@ -81,3 +87,32 @@ export function TransactionProvider({ children }: Props) {
...
+  }
+
+  if (minutesPending > 5) {
+    // every 3 blocks if pending more than 5 minutes
+    return blocksSinceCheck > 2
+  }
+
+  // otherwise every block
+  return true
+}
```
marekkirejczyk, line 118: Could use a test here :P

marekkirejczyk (approved):

PR #110, opened 2021-03-31
🎤 Create notifications example

nezouse:


packages/example/src/components/Notifications/Forms.tsx
```diff
@@ -0,0 +1,109 @@
...
+const wethInterface = new utils.Interface(abi)
+
+interface TransferFormProps {
+  balance: BigNumber | undefined
+  send: (value: BigNumber) => void
+  title: string
+  ticker: string
+}
+
+const TransferForm = ({ balance, send, title, ticker }: TransferFormProps) => {
```
marekkirejczyk, line 27: TransactionForm

packages/example/src/components/Notifications/History.tsx
```diff
@@ -0,0 +1,201 @@
...
+    <Table>{children}</Table>
+  </ContentBlock>
+)
+
+interface TransactionListProps {
+  chainId: ChainId
+}
+
+export const TransactionsList = ({ chainId }: TransactionListProps) => {
+  const { transactions } = useTransactionsContext()
```
marekkirejczyk, line 25: Why not just `useTransactions()`

packages/example/src/components/Notifications/History.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const chainTransactions = transactions[chainId] ?? []
+  const reversed = [...chainTransactions].reverse()
```
marekkirejczyk, line 27: Let's do reverse by default (modify reducer).

Idea to discuss:
```ts
useTransactions({chainId: ...})
```
**ChainId**
- chainId is by default null, in which case useTransaction will use useEthers to determine chainId.
- chainId can be single chainId 
- in the future: array of chainIds
- question: how go we show all chainIds? empty array? 0? other?



```ts
{chainId: null }
```

packages/example/src/components/Notifications/History.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+  return (
+    <TableWrapper title="Transactions history">
+      {reversed.map((tx) => (
+        <div key={tx.transaction.hash}>
+          {!tx.receipt && <Spinner />}
+          <span>{tx.transaction.hash.slice(0, 6) + '...' + tx.transaction.hash.slice(-4)}</span>
```
marekkirejczyk, line 34: Use [shortenAddress(...)](https://usedapp.readthedocs.io/en/latest/core.html#shortenaddress)

packages/example/src/components/Notifications/History.tsx
```diff
@@ -0,0 +1,201 @@
...
+interface NotificationPanelProps {
+  title: string
+  icon: ReactNode
+  transactionHash: string
+}
+
+const NotificationPanel = ({ title, icon, transactionHash }: NotificationPanelProps) => {
+  function getEtherscanLink(hash: string) {
+    return `https://kovan.etherscan.io/tx/${hash}`
+  }
```
marekkirejczyk, line 50: Use [getExplorerAddressLink](https://usedapp.readthedocs.io/en/latest/core.html#getexploreraddresslink)

packages/example/src/components/Notifications/History.tsx
```diff
@@ -0,0 +1,201 @@
...
+    </NotificationWrapper>
+  )
+}
+
+interface NotificationItemProps {
+  notification: Notification
+}
+
+const NotificationItem = ({ notification }: NotificationItemProps) => {
+  switch (notification.type) {
```
marekkirejczyk, line 69: You should be able to do a switch/mapping just on the icon, i.e.
```ts
const label = {'transactionStarted': "Transaction started", ....}
...

function NotificationIcon() {
  switch(type) {
    case 'transactionStarted': return (<ClockIcon />)
    ...
  }
}
```

packages/example/src/components/Notifications/History.tsx
```diff
@@ -0,0 +1,201 @@
...
+    default:
+      return null
+  }
+}
+interface NotificationsListProps {
+  chainId: ChainId
+}
+
+export const NotificationsList = ({ chainId }: NotificationsListProps) => {
+  const { notifications } = useNotificationsContext()
```
marekkirejczyk, line 103: Why not just `useNotifications`?

packages/example/src/components/Notifications/History.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const chainNotifications = notifications[chainId] ?? []
+  const reversed = [...chainNotifications].reverse()
```
marekkirejczyk, line 105: Ascending order by default.

packages/example/src/components/TopBar.tsx
```diff
@@ -27,6 +27,10 @@ export function TopBar() {
               {' '}
               Tokens{' '}
             </HeaderLink>
+            <HeaderLink activeClassName="active-page" to="/notifications">
```
marekkirejczyk, line 30: Let's call it Transactions instead of Notifications

marekkirejczyk (changes requested):
Looks very good. See proposed changes in the comments.

packages/example/src/components/Notifications/History.tsx
nezouse: I don't know if I understand this correctly, but anyway I need to check the type of transaction just to get hash from it.
The only thing that I can get from notification by default right now is `type`.

packages/example/src/components/Notifications/History.tsx
marekkirejczyk: ahhh, shall we add shortenTransaction equvalent?

nezouse: We can just remove checking if the address is correct (and checksumming it). Anyway, it should just work on any string which is long enough.

packages/example/src/components/Notifications/History.tsx
```diff
@@ -0,0 +1,201 @@
...
+    case 'transactionSucceed':
+      return (
+        <NotificationPanel
+          title="Transaction succeed"
+          icon={<CheckIcon />}
+          transactionHash={notification.transaction.hash}
+        />
+      )
+    default:
+      return null
```
marekkirejczyk, line 105: Next PR: Add `walletConnected`.

marekkirejczyk (approved):

PR #111, opened 2021-04-01
🪞 Make notifications and transactions reducers more similar

nezouse:


packages/core/src/providers/transactions/provider.tsx
```diff
@@ -54,10 +56,12 @@ export function TransactionProvider({ children }: Props) {
...
-              submittedAt: Date.now(),
-              transaction: tx.transaction,
-              receipt,
+              notification: {
+                type,
+                submittedAt: Date.now(),
+                transaction: tx.transaction,
+                receipt,
+              },
               chainId,
```
marekkirejczyk, line 65: Turns out we don't need chainId after all (it is in transaction already).

packages/core/src/providers/transactions/provider.tsx
```diff
@@ -22,19 +22,19 @@ export function TransactionProvider({ children }: Props) {
   }, [transactions])
 
   const addTransaction = useCallback(
-    (transaction: TransactionWithChainId) => {
+    ({ transaction, chainId }: AddTransactionPayload) => {
       dispatch({
         type: 'ADD_TRANSACTION',
         transaction,
```
marekkirejczyk, line 28: transactionItem

packages/core/src/providers/notifications/model.ts
```diff
@@ -28,7 +28,10 @@ interface WalletConnected {
 
 export type Notification = TransactionStarted | TransactionSucceed | TransactionFailed | WalletConnected
 
-export type NotificationWithChainId = Notification & { chainId: ChainId }
+export interface AddNotificationPayload {
+  notification: Notification
+  chainId: ChainId
```
marekkirejczyk, line 33: Move to Notification `type: 'WalletConnected'`

packages/core/src/providers/transactions/reducer.ts
```diff
@@ -1,29 +1,36 @@
...
+    case 'ADD_TRANSACTION': {
+      const { chainId } = action.payload.transaction
+
+      if (isChainId(chainId)) {
+        return {
+          ...state,
+          [chainId]: (state[chainId] ?? []).concat(action.payload),
+        }
+      } else {
+        throw TypeError('Unsupported chain')
```
marekkirejczyk, line 31: Unsupported -> Unknown

marekkirejczyk (approved):

PR #113, opened 2021-04-01
🥦 Use useTransactions & useNotifications in example

nezouse:
depends on #112


packages/core/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,20 @@
+import { useMemo } from 'react'
+import { useNotificationsContext } from '../providers'
+import { useEthers } from './useEthers'
+
+export function useNotifications() {
+  const { chainId, account } = useEthers()
+  const { addNotification, notifications } = useNotificationsContext()
```
marekkirejczyk, line 7: I think we can inline `useNotificationsContext` here

marekkirejczyk (approved):

PR #142, opened 2021-04-07
Connect contract to signer only in send

nezouse:


marekkirejczyk (approved):

marekkirejczyk (approved):

bitfede:
Hey Guys! not sure if this is related, but I'm not able to call a contract function because I get this error:

Screen Shot 2021-05-23 at 10 58 08 PM

And here a more detailed output:

Screen Shot 2021-05-23 at 11 56 09 PM

I literally followed your documentations step by step, I defined contract and everything else exactly like it's shown in the docs.. It should work but it does not :(

Signer should be an **optional** variable in Transaction... Why is it interrupting the function execution? Where do I even look for Signer? Why is all this not explained in the docs? There is only a link to the ethersjs documentation but I have no clue how that ethers object fits into the useDapp framework :/ 

We have to update the docs asap, can you please provide an example on how to call a contract function?

PR #168, opened 2021-04-15
Add transaction name to transactions and notifications

nezouse:


packages/core/src/providers/notifications/model.ts
```diff
@@ -2,9 +2,19 @@ import { TransactionReceipt, TransactionResponse } from '@ethersproject/abstract
 import { ChainId } from '../../constants'
 
 type NotificationPayload = { submittedAt: number } & (
-  | { type: 'transactionStarted'; transaction: TransactionResponse }
-  | { type: 'transactionSucceed'; transaction: TransactionResponse; receipt: TransactionReceipt }
-  | { type: 'transactionFailed'; transaction: TransactionResponse; receipt: TransactionReceipt }
+  | { type: 'transactionStarted'; transaction: TransactionResponse; transactionName?: string }
```
marekkirejczyk, line 5: Why in one line?

marekkirejczyk (approved):

packages/core/src/providers/notifications/model.ts
nezouse: When they are small it's just easier to compare the types.

PR #192, opened 2021-05-07
Fix out of order requests

sz-piotr:
When multicall requests made for the same block but for different chain calls resolve out of order the data gets overwritten. This PR fixes this by merging state in such cases.

### DIFF
--- a/packages/core/src/providers/chainState/chainStateReducer.ts
+++ b/packages/core/src/providers/chainState/chainStateReducer.ts
@@ -1,6 +1,6 @@
 import { ChainState } from './model'
 
-interface State {
+export interface State {
   [chainId: number]:
     | {
         blockNumber: number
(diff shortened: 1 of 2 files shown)


packages/core/src/providers/chainState/chainStateReducer.ts
```diff
@@ -30,9 +30,24 @@ export function chainStateReducer(state: State = {}, action: Action) {
   const current = state[action.chainId]?.blockNumber
   if (!current || action.blockNumber >= current) {
     if (action.type === 'FETCH_SUCCESS') {
+      let newState = action.state
+      if (action.blockNumber === current) {
+        // merge with existing state to prevent requests coming out of order
+        // from overwriting the data
+        const oldState = state[action.chainId]?.state ?? {}
```
marekkirejczyk, line 37: Could extract `mergeChainData`  or st

PR #198, opened 2021-05-12
Refactor send transaction example

Szymx95:


packages/example/src/components/Transactions/TransactionForm.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+const getTransactionErrorMessage = (transaction: TransactionStatus) => {
+  if ('errorMessage' in transaction) return transaction.errorMessage
+  else return ''
+}
+
+const getTransactionChainId = (transaction: TransactionStatus) => {
+  if ('chainId' in transaction) return transaction.chainId
+  else return undefined
+}
```
marekkirejczyk, line 53: No need to extract helper, just use `transaction?.chainId`

packages/example/src/components/Transactions/TransactionForm.tsx
```diff
@@ -0,0 +1,285 @@
...
+    <div style={{ color: color, textAlign: 'center' }}>{text}</div>
+  </InformationRow>
+)
+
+const transactionErrored = (transaction: TransactionStatus) => {
+  return 'errorMessage' in transaction
+}
+
+const getTransactionErrorMessage = (transaction: TransactionStatus) => {
+  if ('errorMessage' in transaction) return transaction.errorMessage
```
marekkirejczyk, line 46: No need to extract helper, just use `transaction?.errorMessage`

marekkirejczyk: See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html

nezouse: It's correct this way, TranscationStatus without checks is just 
```js
{
      status: string
}
```

marekkirejczyk: Fair enough :)
I was wondering if we can do:
```ts
transactionErrored(transaction) && transaction?.errorMessage
```
Does not seem awesome.

marekkirejczyk (approved):

PR #202, opened 2021-05-13
Set TransactionStatus as interface and move it to model

Szymx95:


packages/core/src/hooks/useContractFunction.ts
```diff
@@ -72,21 +44,21 @@ export function useContractFunction(contract: Contract, functionName: string, op
 
     try {
       transaction = await contractWithSigner[functionName](...args)
-      setState({ transaction, status: 'Mining', chainId })
+      setState({ transaction, status: 'Mining', errored: false, chainId })
```
noobshow, line 47: Does errored seem necessary?
I'm thinking that's why errorMessage can be undefined, so users can just check if(errorMessage) //there's an error

packages/core/src/hooks/useContractFunction.ts
```diff
@@ -72,21 +44,21 @@ export function useContractFunction(contract: Contract, functionName: string, op
...
-      setState({ receipt, transaction, status: 'Success', chainId })
+      setState({ receipt, transaction, status: 'Success', chainId, errored: false })
     } catch (e) {
       const errorMessage = e.reason ?? e.message
       if (transaction) {
-        setState({ status: 'Fail', transaction, receipt: e.receipt, errorMessage, chainId })
+        setState({ status: 'Fail', transaction, receipt: e.receipt, errorMessage, chainId, errored: true })
       } else {
-        setState({ status: 'Exception', errorMessage, chainId })
+        setState({ status: 'Exception', errorMessage, chainId, errored: true })
```
noobshow, line 61: You noticed there's an errorMessage value defined there, but not in other places? Because it's meant to show same thing you proposed {errored:true}

packages/example/src/components/Transactions/TransactionForm.tsx
```diff
@@ -38,20 +38,6 @@ const StatusBlock = ({ color, text, icon }: StatusBlockProps) => (
...
-const transactionErrored = (transaction: TransactionStatus) => {
-  return 'errorMessage' in transaction
-}
-
-const getTransactionErrorMessage = (transaction: TransactionStatus) => {
-  if ('errorMessage' in transaction) return transaction.errorMessage
-  else return ''
-}
-
-const getTransactionChainId = (transaction: TransactionStatus) => {
```
noobshow, line 50: This seems to be useful or might be useful, not sure how others view it, but same value can be gotten directly without a shorthand function...

Maybe the shorthand can be left to users?

noobshow (commented):
Moving the transactionStatus into model seems good

packages/core/src/model/TransactionStatus.ts
```diff
@@ -0,0 +1,11 @@
...
+import { TransactionResponse, TransactionReceipt } from '@ethersproject/abstract-provider'
+
+export interface TransactionStatus {
+  status: 'None' | 'Mining' | 'Success' | 'Fail' | 'Exception'
+  errored: boolean
+  transaction?: TransactionResponse
+  receipt?: TransactionReceipt
+  chainId?: ChainId
+  errorMessage?: string
+}
```
nezouse, line 11: With `errored` property you can now do discriminated union like:
```ts
{errored: true; errorMessage: string } | {errored: false; errorMessage: never}
```

noobshow: Yea, just my own opinion... They looks same to me

noobshow: All I'm saying is without the boolean, the same could still be achieved...

noobshow: And, with just one way, there's surely a unified approach than having multiple variables doing same thing, it gives dev users a kind of fear both don't work the same way or they meant something different from development perspective...

Might just be a bad opinion rather

packages/core/src/hooks/useContractFunction.ts
marekkirejczyk: How about extracting:
```ts
function transactionErrored(transaction) {
  return ...
}
```

packages/core/src/model/TransactionStatus.ts
marekkirejczyk: After some discussion, I think it is fine approach (the whole type safety approach is confusing for many people and leads to hacking code every now an then). Let describe it correctly in changeset and PR name

packages/core/src/model/TransactionStatus.ts
```diff
@@ -0,0 +1,11 @@
+import { ChainId } from '../constants'
+import { TransactionResponse, TransactionReceipt } from '@ethersproject/abstract-provider'
+
+export interface TransactionStatus {
+  status: 'None' | 'Mining' | 'Success' | 'Fail' | 'Exception'
```
marekkirejczyk, line 5: Extract type `TransactionState`?

packages/core/src/hooks/useContractFunction.ts
noobshow: You mean exposing it to the public as well?

My opinion might not matter, but the reason I see this project as a life saver worth exploring for usage is how it doesn't expose much any unnecessary things, keeping it lightweight, yet adding major features that makes using dapps through hooks viable... I don't know if the goal is different

noobshow: Is errored the new term for failed?

packages/core/src/model/TransactionStatus.ts
noobshow: My argument is not against the boolean added, more against duplication of errorMessage, which in turn means everywhere TransactionStatus is initialized, you have to set errored|failed to false instead of just not bothering and not initializing errorMessage

packages/core/src/model/TransactionStatus.ts
```diff
@@ -1,11 +1,16 @@
...
-  errored: boolean
+  status: TransactionState
   transaction?: TransactionResponse
   receipt?: TransactionReceipt
   chainId?: ChainId
   errorMessage?: string
 }
+
+export function transactionErrored(transaction: TransactionStatus) {
+  return 'errorMessage' in transaction
```
noobshow, line 15: This seems good

packages/core/src/hooks/useContractFunction.ts
noobshow: Sorry, I misquote your https://github.com/EthWorks/useDApp/pull/202#discussion_r632434360 to extract the errored to function... Seems good now

packages/core/src/model/TransactionStatus.ts
nezouse: This function is basically noop now, errorMessage is always in TransactionStatus.

nezouse: A better way to check it now would be `transaction.errorMessage !== undefined`, but both works.

noobshow: @nezouse 
True about the helper, "one helper, more helper"... 

It might be just good to allow the users define helpers in a way it fits their various projects, coz some would even eventually still provide a helper around this helper because they want to use it in a class or library they use everywhere in their app...

This project is cool being simple, yet extracting the hardwork of interfacing with ethers.js directly to a hook...

marekkirejczyk: Fair enough. But one don't need to use helper :P

marekkirejczyk (approved):

noobshow:
Leaving the transactionErrored helper in model?

noobshow (commented):
All good, but Helper not really necessary

marekkirejczyk:
We left the helper. But it is no necessary to use it :)

PR #204, opened 2021-05-14
Move core tests to packages/core/test

Szymx95:


packages/testing/src/renderWeb3Hook.tsx
```diff
@@ -36,11 +35,7 @@ export const renderWeb3Hook = async <Tprops, TResult>(
   const { result, waitForNextUpdate, rerender, unmount } = renderHook<Tprops, TResult>(hook, {
     wrapper: (wrapperProps) => (
       <MockWeb3Wrapper connector={connector}>
-        <BlockNumberProvider>
-          <ChainStateProvider multicallAddresses={multicallAddresses}>
-            <UserWrapper {...wrapperProps} />
-          </ChainStateProvider>
-        </BlockNumberProvider>
```
nezouse, line 43: We can't just remove core features to make something look better. Without that wrapper, we can't test any of the hooks calling the contracts.

nezouse (changes requested):

packages/testing/src/renderWeb3Hook.tsx
Szymx95: That's why i think tests should provide wrappers of providers, because some tests might need different wrappers

packages/testing/examples/useBlockMeta.test.tsx
```diff
@@ -1,10 +1,25 @@
...
+      <BlockNumberProvider>
+        <ChainStateProvider multicallAddresses={multicallAddresses}>{children}</ChainStateProvider>
+      </BlockNumberProvider>
+    )
+    webHookOptions = { mockProvider, mockConnector, renderHook: { wrapper } }
+  })
+
   it('retrieves block timestamp and difficulty', async () => {
-    const { result, waitForCurrent } = await renderWeb3Hook(useBlockMeta)
+    const { result, waitForCurrent } = await renderWeb3Hook(useBlockMeta, webHookOptions)
```
Szymx95, line 23: I think tests should be providing wrappers, i know its a major change in functionality but in my opinion it will make testing better in the future.

Just a opinion though.

nezouse: Sure, but you could add custom wrappers before.  Now you just need to add them by yourself in every single test. And the point of the testing library is to reduce that boilerplate.

Szymx95: I think all the tests should be in one place so development in the future will be easier and less confusing.

Szymx95: So other option is to move all the tests to something like /packages/testing/testCore

Szymx95: Also MockBlockNumberProvider could be created as it's a simple provider also creation of mock provider,connector and multicall addresses could be moved to testing library so there is still some place for refactor.

I could commit some refactoring before this pull request is merged, but i think it's a good start

packages/core/test/hooks/useBlockNumber.test.tsx
```diff
@@ -1,22 +1,13 @@
...
-    const mockProvider = new MockProvider()
-    const mockConnector = new MockConnector(mockProvider)
-    const multicallAddresses = await deployMulticall(mockProvider, mockConnector, MultiCall)
-    const wrapper: React.FC = ({ children }) => (
-      <BlockNumberProvider>
-        <ChainStateProvider multicallAddresses={multicallAddresses}>{children}</ChainStateProvider>
-      </BlockNumberProvider>
-    )
-    webHookOptions = { mockProvider, mockConnector, renderHook: { wrapper } }
+    webHookOptions = await createChainStateProviderHookOptions(BlockNumberProvider, ChainStateProvider, MultiCall)
```
Szymx95, line 10: Moved boiler plate to another function. Better name for function is welcome.

packages/core/package.json
```diff
@@ -20,6 +20,7 @@
     "@typescript-eslint/parser": "^4.10.0",
     "@web3-react/abstract-connector": "^6.0.7",
     "@web3-react/types": "^6.0.7",
+    "@usedapp/testing": "*",
```
marekkirejczyk, line 23: "^..."

marekkirejczyk:
Closing for now

PR #205, opened 2021-05-14
Corrected useTokenBalance() arguments order

antooni:
In Core API hook is documented correctly, only there was an error.

### DIFF
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -137,11 +137,11 @@ Provides a way to fetch balance of ERC20 token specified by ``tokenAddress`` for
 
   export function TokenBalance() {
     const { account } = useEthers()
-    const tokenBalance = useTokenBalance(account, DAI)
+    const tokenBalance = useTokenBalance(DAI, account)
 
     return (
       <div>
-        {tokenBalance && <p>Balance: {formatUnits(tokenBalance, 18)}</p>}
+        {tokenBalance && <p>Balance: {formatUnits(tokenBalance as any, 18)}</p>}
       </div>
     )
   }
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -141,7 +141,7 @@ Provides a way to fetch balance of ERC20 token specified by ``tokenAddress`` for
 
     return (
       <div>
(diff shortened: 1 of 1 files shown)


docs/source/getting-started.rst
```diff
@@ -137,11 +137,11 @@ Provides a way to fetch balance of ERC20 token specified by ``tokenAddress`` for
 
   export function TokenBalance() {
     const { account } = useEthers()
-    const tokenBalance = useTokenBalance(account, DAI)
+    const tokenBalance = useTokenBalance(DAI, account)
```
marekkirejczyk, line 140: 👍

docs/source/getting-started.rst
```diff
@@ -137,11 +137,11 @@ Provides a way to fetch balance of ERC20 token specified by ``tokenAddress`` for
...
 
     return (
       <div>
-        {tokenBalance && <p>Balance: {formatUnits(tokenBalance, 18)}</p>}
+        {tokenBalance && <p>Balance: {formatUnits(tokenBalance as any, 18)}</p>}
```
marekkirejczyk, line 144: Why as any?

antooni: Without it I got an error:
  " Argument of type 'BigNumber' is not assignable to parameter of type 'BigNumberish' "

noobshow: I don't get such error here, do you have another BigNumber library imported in the scope? Maybe other BigNumber library

marekkirejczyk: You should be able to fix it by updating to the newest ethers and useDApp.

noobshow: Sometimes, I get such error... But, I try to check properly and most times saw I'm not doing something right

marekkirejczyk: There was in the older version problem when app was using ethers 5.1.* and useDApp was using ethers 5.0.*
Latest useDApp version is using 5.1.* and so should your app use. With that it should all work and there should be no type mismatches.

marekkirejczyk (changes requested):
Remove `as any`

marekkirejczyk (approved):

marekkirejczyk (approved):

PR #207, opened 2021-05-14
Add useSendTransaction, refactor useContractFunction

Szymx95:


packages/core/src/hooks/useSendTransaction.ts
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }
+}
+
+interface Options {
+  signer?: Signer
+  transactionName?: string
+}
+
+export function useSendTransaction(options?: Options) {
```
marekkirejczyk, line 44: There should be at least on more argument - `transaction: TransactionRequest`, with amount, nonce, etc

packages/core/src/hooks/useSendTransaction.ts
```diff
@@ -0,0 +1,65 @@
+import { TransactionResponse, TransactionRequest } from '@ethersproject/abstract-provider'
+import { Signer } from 'ethers'
+import React, { useState } from 'react'
+import { StoredTransaction, useTransactionsContext } from '../providers'
+import { TransactionStatus } from '../../src'
+import { useEthers } from './useEthers'
+
+async function handleTransactionState(
```
marekkirejczyk, line 8: 👍 Looks like you could use this function in `useContractFunction`

packages/core/src/hooks/useSendTransaction.ts
```diff
@@ -0,0 +1,65 @@
...
+    } else {
+      setState({ status: 'Exception', errorMessage, chainId })
+    }
+  }
+}
+
+interface Options {
+  signer?: Signer
+  transactionName?: string
+}
```
marekkirejczyk, line 42: Let's extract common TransactionHookOptions and use the same for type for `useContractFunction`

marekkirejczyk: We can than reuse `connectContractToSigner` as well in both functions

marekkirejczyk: Idea: Can we extract on more generic function, i.e.:
```ts
useSendGenericTransaction(() => Promise<TransactionResponse>, options?: Options) {
  ...
}
```

So that useSendTransaction would be something like:
```ts
function useSendTransaction(...) {
  return useSendGenericTransaction(() => 
    signer.sendTransaction(transactionRequest)
  )
}
```

packages/testing/examples/useSendTransaction.test.ts
```diff
@@ -0,0 +1,48 @@
...
+import { MockProvider } from 'ethereum-waffle'
+import { BigNumber, utils } from 'ethers'
+import { renderWeb3Hook } from '../src'
+
+describe('useSendTransaction', () => {
+  const mockProvider = new MockProvider()
+  const [spender, receiver, secondReceiver] = mockProvider.getWallets()
+
+  it('success', async () => {
+    const spenderBalance = await receiver.getBalance()
```
marekkirejczyk, line 12: receiver.getBalance() ->  spender.getBalance()

marekkirejczyk (approved):

bitfede:
Hey Guys! not sure if this is related, but I'm not able to call a contract function because I get this error:

Screen Shot 2021-05-23 at 10 58 08 PM

And here a more detailed output:

Screen Shot 2021-05-23 at 11 56 09 PM

I literally followed your documentations step by step, I defined contract and everything else exactly like it's shown in the docs.. It should work but it does not :(

Signer should be an **optional** variable in Transaction... Why is it interrupting the function execution? Where do I even look for Signer? Why is all this not explained in the docs? There is only a link to the ethersjs documentation but I have no clue how that ethers object fits into the useDapp framework :/ 

We have to update the docs asap, can you please provide an example on how to call a contract function?

marekkirejczyk:
Try:
```ts
const {activate, account, library}  = useEthers();
const {state, send}  = useContractFunction(contract, 'addNewOpera', library);
....
```

Szymx95:
Propably wallet is not connected did you try:

```
const { activateBrowserWallet } = useEthers();
activateBrowserWallet()
```

Before using `send` ?

bitfede:
Thanks for your reply @marekkirejczyk 

Unfortunately I still get the error:
Screen Shot 2021-05-24 at 8 26 16 AM

Terminal:
Screen Shot 2021-05-24 at 8 38 03 AM
I see undefined as log output..

I realized that the issue might be that I'm using NextJS, and the code is being run also on server-side.

What could be a nextJS-friendly approach?

----

@Szymx95 thanks for your reply! Good observation, yes I do have the `activateBrowserWallet` in another component

PR #210, opened 2021-05-14
Add useDapp DevTools extension

sz-piotr:
This PR introduces a DevTools extension for working with useDapp.
Screenshot from 2021-05-14 20-15-31

You can find more information in the readme. The extension is working and fully usable. It probably needs more polish before release but I am done working on it.

This was originally intended to be a part of the 0xHack hackaton, but for reasons outside of the scope of this PR it will not be eligible.


noobshow:
My guess is this wouldn't(or shouldn't) be bundled in the @core package since it might affect the size and the extension would be used in development

packages/extension/src/hooks/useStorage.ts
```diff
@@ -0,0 +1,44 @@
+/* global chrome */
+import { useEffect, useState } from 'react'
+
+export function useStorage<T>(key: string) {
+  const [state, setState] = useState<T | undefined>(undefined)
+
+  useEffect(() => {
+    setState(undefined)
+    if (window.chrome) {
```
marekkirejczyk, line 9: if (window.chrome?.storage) {

sz-piotr:
@noobshow Sorry for the late reply. You are correct that including extension code in core is bad. This is why this PR includes the bare minimum glue in core and all of the code is actually independent and placed under the usedapp/extension package.

@marekkirejczyk I have merged the latest changes from master into this branched and manually verified that it still works.

packages/extension/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "lint:prettier": "prettier --check .",
+    "test": "mocha",
+    "storybook": "start-storybook -s ./static -p 6006",
+    "build-storybook": "build-storybook -s ./static"
+  },
+  "dependencies": {
+    "ethers": "^5.1.4",
+    "react": "^17.0.0",
+    "react-dom": "^17.0.0",
+    "react-router-dom": "^5.2.0",
```
marekkirejczyk, line 21: I don't think react-router-dom is actually used.

packages/extension/src/hooks/useStorage.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export function useStorage<T>(key: string) {
+  const [state, setState] = useState<T | undefined>(undefined)
+
+  useEffect(() => {
+    setState(undefined)
+    if (window.chrome?.storage) {
+      let cancelled = false
+      chrome.storage.local.get([key], function (result) {
+        console.log(`STORAGE ${key}:`, result)
```
marekkirejczyk, line 12: console.log?

packages/extension/src/providers/abi/AbiEntry.ts
```diff
@@ -0,0 +1,24 @@
+import { Interface, FunctionFragment, FormatTypes } from '@ethersproject/abi'
+
+export interface AbiEntry {
+  code: string
+  selector: string
+  coder: Interface
+  fragment: FunctionFragment
+}
+
+export function toAbiEntry(abi: any): AbiEntry | undefined {
```
marekkirejczyk, line 10: `abi:Abi`
`type Abi = string[]`?

packages/extension/src/views/Events/EventItem/formatInteger.ts
```diff
@@ -0,0 +1,12 @@
+export function formatInteger(integer: number | string): string {
+  const value = integer.toString()
```
marekkirejczyk, line 2: Why not simply:
`new Intl.NumberFormat('en-US').format(1000000)`

Currently:
`formatInteger('-10.03') === "-10,.03"`

packages/extension/src/views/Events/EventPreview/BlockFoundPreview.tsx
```diff
@@ -0,0 +1,47 @@
...
+          <Property name="Explore">
+            <Link href={link} />
+          </Property>
+        )}
+      </Table>
+    </>
+  )
+}
+
+function getExplorerLink(network: string, blockNumber: number) {
```
marekkirejczyk, line 32: 😢 Duplication with core, I know I know, extracting separate package, e.t.c.

marekkirejczyk (approved):

packages/extension/src/views/Events/EventItem/formatInteger.ts
sz-piotr: You can't use Intl.NumberFormat because the numbers can be outside of the safe range of 2^53

sz-piotr: Also `formatInteger` formats integers not floats :)

PR #211, opened 2021-05-16
Add BSC support

gasolin:
Add BSC chain id, explorer
and the multicall contract on https://github.com/makerdao/multicall/pull/20/files

test, lint passed

### DIFF
--- a/.changeset/tiny-teachers-approve.md
+++ b/.changeset/tiny-teachers-approve.md
@@ -1,0 +1,5 @@
+---
+"@usedapp/core": patch
+---
+
+Add BSC support
--- a/docs/source/core.rst
+++ b/docs/source/core.rst
@@ -456,7 +456,7 @@ ChainId
 Enum that represents chain ids.
 
 **Values:**
-``Mainnet, Goerli, Kovan, Rinkeby, Ropsten, xDai``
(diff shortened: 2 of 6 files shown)


noobshow:
I'm not sure all these makes bsc support work fully, you've tested it on metamask?

You know you still need some connectors like this :

https://github.com/binance-chain-npm/bsc-web3-connector

See an example usage here https://github.com/aragon/use-wallet/tree/master/examples/binance-chain that uses the same web3-react used in this same library

marekkirejczyk (approved):

marekkirejczyk:
I think it should make it work with metamask. binnance wallet will require connector.

gasolin:
@marekkirejczyk thanks for the merge!

For the future version, is it make sense to create separate packages for different EVM compatible chains (ex public bsc related thing as @usedapp/bsc, @usedapp/xdai in packages/) and only put main Ethereum and reusable functions in `@usedapp/core`?

marekkirejczyk:
Thanks for your work ❤️ 

Yes it makes sense to only put main Ethereum and reusable functions in `@usedapp/core` and the custom code per chain (eg bsc ) to separate  packages.

Chain ids and basic functionality (like explorer links) are fine in `core`. I would not consider them "custom code", just configuration. If you would like to add connector -> I would create new package for that.

PR #218, opened 2021-05-18
Add useSlowTransactions hook

Szymx95:


packages/core/src/helpers/transaction.ts
```diff
@@ -15,3 +16,10 @@ export function shortenIfTransactionHash(transactionHash: string | Falsy): strin
   }
   return ''
 }
+
+export function isTransactionSlow(transaction: StoredTransaction, slowTransactionThreshold: number): boolean {
```
marekkirejczyk, line 20: Idea:
```ts
const slow = useTransactionSlow(transactionHash: string) 
```

That way we can:
- check if transaction is slow regardless if it is a stored transaction or just transaction
- use useConfig inside isTransactionSlow

packages/core/src/hooks/useSlowTransactions.ts
```diff
@@ -0,0 +1,70 @@
...
+  setSlowTransactions: React.Dispatch<React.SetStateAction<{ hash: string; submittedAt: number }[]>>,
+  transactions: { hash: string; submittedAt: number }[],
+  threshhold: number,
+  library: Web3Provider
+) {
+  const results = await Promise.all(
+    transactions.map(async ({ hash }) => {
+      return await library?.getTransaction(hash)
+    })
+  )
```
nezouse, line 18: Requesting transactions one by one will probably be safer. Right now it can start nearly an unlimited amount of web requests at once.  
`map` is not working well with async code so it's not waiting for the request to finish before starting the next one.

packages/core/src/hooks/useSlowTransactions.ts
```diff
@@ -0,0 +1,70 @@
...
+        if ('blockHash' in results[index]) {
+          return false
+        }
+      }
+      return true
+    })
+  )
+  setSlowTransactions(
+    transactions.filter(({ submittedAt }) => {
+      if (Date.now() - submittedAt > threshhold) {
```
nezouse, line 32: Why not just `return Date.now() - submittedAt > threshold`?

packages/testing/examples/useSlowTransactions.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+    await result.current.sendTransaction({})
+
+    expect(result.current.slowTransactions.length).to.eq(2)
+    expect(result.current.slowTransactions[0].hash).to.eq(hashes[0])
+    expect(result.current.slowTransactions[1].hash).to.eq(hashes[1])
+
+    await sleep(300)
+    await result.current.sendTransaction({})
+    await sleep(500)
```
nezouse, line 50: Please try to use `waitForCurrent` instead of fixed sleep if it's possible.

marekkirejczyk:
With no clear path to implement transaction speed-up, we are closing this for now.

packages/core/src/helpers/transaction.ts
noobshow: Or it could check both Typed Transaction or hash of such transactions through :

`transaction: StoredTransaction | string`

When its a string, expect it to be a hash and when its StoredTransaction, check as usual...

PR #219, opened 2021-05-18
Add transaction overrides

Szymx95:


packages/core/src/hooks/useTransactionOverride.ts
```diff
@@ -0,0 +1,72 @@
...
+  const { promiseTransaction, state } = usePromiseTransaction(chainId)
+
+  const overrideTransaction = async (
+    transaction: StoredTransaction,
+    transactionRequest: TransactionRequest
+  ): Promise<overrideRespond> => {
+    const request: TransactionRequest = { ...transactionRequest, nonce: transaction.transaction.nonce }
+    if (library?.connection.url === 'metamask') {
+      return { request, error: "Metamask doesn't support custom nonce" }
+    }
```
Szymx95, line 23: Metamask doesn't support custom nonce injection so to avoid doubling transactions function returns without sending

packages/core/src/providers/transactions/model.ts
```diff
@@ -7,6 +8,7 @@ export interface StoredTransaction {
   receipt?: TransactionReceipt
   lastCheckedBlockNumber?: number
   transactionName?: string
+  signer?: Signer
```
Szymx95, line 11: signer added so that overriding transaction is sent from the same account

packages/testing/examples/useTransactionOverride.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+      expect(overrideResult.request.gasPrice).to.eq(result.current.state.transaction?.gasPrice.add('10000000000'))
+      expect(overrideResult.request.value).to.eq(0)
+    })
+  })
+
+  describe('speedUpTransaction', () => {
+    it('success', async () => {
+      const { result, waitForCurrent } = await renderWeb3Hook(
+        () => {
+          const { sendTransaction, state } = useSendTransaction()
```
marekkirejczyk, line 43: I was thinking about a bit different API:
```ts
const { sendTransaction, state, speedUp, cancel } = useSendTransaction()
```

packages/core/src/hooks/useTransactionOverride.ts
```diff
@@ -0,0 +1,72 @@
...
+        transaction.transactionName
+      )
+      return { request }
+    } else {
+      return { request, error: 'Transaction has no signer' }
+    }
+  }
+
+  const cancelTransaction = async (transaction: StoredTransaction, gasIncrease?: BigNumber) => {
+    const gasTxIncrease = gasIncrease || BigNumber.from('10000000000')
```
marekkirejczyk, line 37: gasIncrease has to be at least a 10% higher gas fee (protocol rule)

marekkirejczyk:
No supported by MetaMask, closing for now :(

PR #222, opened 2021-05-20
Change notifications to popups

Szymx95:
Screenshot 2021-05-21 at 10 40 37


marekkirejczyk:
- I would add more info about transaction (i.e. shorten transaction hash)
- Let's not make items transparent

marekkirejczyk (approved):

PR #227, opened 2021-05-22
[PR for feedback] add useTokenPrice hook

gasolin:
- add useTokenPrice hook,  given token and quote to get token price
- show Ether price in example (Balance)

Imgur

- allow pass different adapter to get token price
- current the CoinGecko adapter is implemented without dependency https://www.coingecko.com/en/api

TODO
- [ ] auto update while block update
- [ ] document
- [ ] test


packages/core/src/hooks/useTokenPrice.ts
```diff
@@ -0,0 +1,36 @@
...
+  useEffect(() => {
+    async function getPrice() {
+      const tokenPrice = await getTokenPrice(token, quote, adapter)
+      setPrice(tokenPrice)
+    }
+
+    if (token && quote) {
+      getPrice()
+    }
+  }, [token, quote])
```
noobshow, line 16: add `adapter` to the deps array

packages/core/src/hooks/useTokenPrice.ts
```diff
@@ -0,0 +1,19 @@
+import { useState, useEffect } from 'react'
+
+import { getTokenPrice, adapterCoinGecko } from '../helpers/price'
+
+export const useTokenPrice = (token: string, quote: string, adapter = adapterCoinGecko) => {
```
Szymx95, line 5: since useTokenBalance accepts a address of a token i feel like useTokenPrice should also take address instead of token id

packages/core/src/helpers/price.ts
```diff
@@ -0,0 +1,18 @@
...
+      'Content-Type': 'application/json',
+    },
+  })
+  // return format {"ethereum":{"usd":2234.6}}
+  const result = await data.json()
+  return result[tokenId][quoteId]
+}
+
+export async function getTokenPrice(token: string, quote: string, adapter = adapterCoinGecko) {
+  return await adapter(token, quote)
```
Szymx95, line 17: Maybe wrap this inside try/catch block

packages/core/src/hooks/useTokenPrice.ts
noobshow: Possibly, there should be a way to map token address(or symbol) to the token_id required to interact with coingecko...

marekkirejczyk:
Hi @gasolin,

### Feedback

Great work so far ❤️ 

We discussed it extensively with @Szymx95 and we propose following improvements:

- Let's do just one function `getCoingGeckoPrice(...)` (currently `adapterCoinGecko`) and skip the whole adapter thing


### Additional ideas for extension (optional):

- Adding `useCoinGeckoPrice(...)` that would refresh on every new block, ie
```ts
function useCoinGeckoPrice(...) {
  const blockNo = useBlockNumber();
  const [price, setPrice] = useState()
  useEffect(() => { 
    setPrice(getCoingGeckoPrice(...));
  }, [blockNo])
}
```

- Adding `getCoingGeckoPrice` and `useCoinGeckoTokenPrice` that uses address instead of coin name (and uses this API: `/simple/token_price/{id}`)

- Extract to separate package @usedapp/coingecko

gasolin:
thanks for the feedback!

PR #230, opened 2021-05-23
Add useGasPrice hook

flyinglimao:
This PR introduces a new hook `useGasPrice(pollingInterval?: number) => number`, which will get the gas price via `Web3Provider.perform('getGasPrice', null)` and poll new value according to `pollingInterval` passed or config (library.pollingInterval).
The reason for providing a parameter to set a standalone interval is that for some DApp, polling new data isn't important (e.g. NFT wallet) but knowing fee is (for sending).
It might be useful for computing fee or knowing status of chains


packages/core/src/hooks/useGasPrice.ts
```diff
@@ -0,0 +1,20 @@
+import { useEffect, useState } from 'react'
+import { useEthers } from './useEthers'
+
+export function useGasPrice(pollingInterval?: number): number {
+  const { library } = useEthers()
+  const [gasPrice, setGasPrice] = useState(0)
+
+  useEffect(() => {
+    async function updateGasPrice() {
+      const newGasPrice: string = await library?.perform('getGasPrice', null)
```
Szymx95, line 10: Why not `library?.getGasPrice()` and return a BigNumber instead of number

flyinglimao: Got error in my ide so I thought it wasn't provided but it shows up now. Updated

marekkirejczyk:
Hi @flyinglimao,

Great work so far ❤️ 

### Proposal

We researched and discussed with @Szymx95 and here is what we come up with:

1. If you want to check the price just once, you can simply use `await library?.getGasPrice()`

2. If you want to have a current price, the gas estimation changes on every block (tested with Infura), so it would make sense to check on every new block, i.e.:

```ts
function useGasPrice(...) {
  ...
  const blockNo = useBlockNumber();
  const [gasPrice, setGasPrice] = useState()
  useEffect(() => { 
    setGasPrice(await library?.getGasPrice());
  }, [blockNo])
}
```

### Optional

In the future we could optimise useGasPrice some more, so it only triggers once per block, even if used in multiple places (and never if not used). For that we would need a context with reference counting and debouncing. That would be very similar to `useContactCall`.

marekkirejczyk:
Looks great! Would add a simple description in docs?

Szymx95:
Hi,

awesome work,

I have an idea maybe ``useGasPrice()`` should return ``BigNumber | undefined`` with initial gas price state being ``undefined``

```ts
const [gasPrice, setGasPrice] = useState(undefined)
```

In order to avoid reading gasPrice as 0 before it is initialised with different value

packages/testing/examples/useGasPrice.test.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+chai.use(solidity)
+
+describe('useGasPrice', () => {
+  it('retrieves gas price', async () => {
+    const { result, waitForCurrent } = await renderWeb3Hook(useGasPrice)
+    await waitForCurrent((val) => val !== undefined)
+
+    expect(result.error).to.be.undefined
+    expect(result.current.toNumber()).to.be.a('number')
```
Szymx95, line 14: I think it should be like this:
```ts
expect(result?.current?.toNumber()).to.be.a('number')
```
Otherwise typescript will say that object is possibly undefined.

flyinglimao: Updated, `result` is expected not to be undefined therefore I used `result.current?.toNumber()`

marekkirejczyk (approved):

PR #233, opened 2021-05-25
Add wallet activation error handling

Szymx95:


packages/core/src/hooks/useEthers.ts
```diff
@@ -14,9 +14,16 @@ export type Web3Ethers = ReturnType<typeof useWeb3React> & {
...
-    return () => result.activate(injected)
-  }, [supportedChains])
+  const activateBrowserWallet = useCallback(
+    async (onError?: (error: Error) => void, throwErrors?: boolean) => {
+      const injected = new InjectedConnector({ supportedChainIds: supportedChains })
+      if (onError instanceof Function) {
+        await result.activate(injected, onError, throwErrors)
+      } else {
+        await result.activate(injected, undefined, throwErrors)
+      }
```
Szymx95, line 24: onError needs to be checked if its a function so that acitvateBrowserWallet can be passed to onClick like
```ts
<button onClick={avticateBrowserWallet} />
```

marekkirejczyk (approved):

PR #240, opened 2021-05-25
add @usedapp/coingecko package

gasolin:
replace #227

- add `useCoingeckoPrice` hook in `@usedapp/coingecko`,  given base and quote to get token price
- add `useCoingeckoTokenPrice` hook, given the token contract to get token price
- show Ether, WETH price in example (Balance)
- price get update while a new block is mined
- base API test and doc included

PR #241, opened 2021-05-26
Add @usedapp/chainlink which provides useChainlinkPrice hook

flyinglimao:
Chainlink provides a way to get price which not dependent on centralized services. This PR introduces a hook `useChainlinkPriceFeed(oracleAddress: string) => number | undefined`.
This hook will turn raw data `latestAnswer` and `decimals` to `price` which is a native number to simplify usage. (I think that most case will only need price instead of those two value).
ABI was cropped from @chainlink/abi

The mock file used in test was published on IPFS
```
ChainlinkPriceFeedMock.sol : 
dweb:/ipfs/QmZbF52Bpkp68YKJMdxbBCtF7YHqkppu3u1Qq32WrzatKz

metadata.json : 
dweb:/ipfs/QmeBdgC8mkMLn3h4qBkpba3TQaz5uPjqWQDGwPnmvanhAj
```


noobshow:
While this seems cool, your implementation seems complicated...

It could be made easier, some calls can be chained together too...

Maybe you follow this https://github.com/EthWorks/useDApp/pull/240 structure to make sure the implementations follow a standard?

flyinglimao:
> While this seems cool, your implementation seems complicated...
> 
> It could be made easier, some calls can be chained together too...
> 
> Maybe you follow this #240 structure to make sure the implementations follow a standard?
> 
> And useChainLinkPrice() might be a better name, maybe a separate package too like in the other implementation linked, return price only...

I don't know why you say my implementation is complicated. The reason not using `useChainCalls` is that it need more code and conditional codes or maybe I didn't find a good way to use it.
The PR you mentioned didn't say any word about a standard, so I don't know what standard to follow. Besides, there exists a `useEtherPrice` in roadmap. Therefore I think it's fine to put it in core but maybe later there will be another package like `@usedapp/price` providing price hooks from Chainlink, CoinGecko, or even Binance, FTX.

noobshow:
> > While this seems cool, your implementation seems complicated...
> > It could be made easier, some calls can be chained together too...
> > Maybe you follow this #240 structure to make sure the implementations follow a standard?
> > And useChainLinkPrice() might be a better name, maybe a separate package too like in the other implementation linked, return price only...
> 
> I don't know why you say my implementation is complicated. The reason not using `useChainCalls` is that it need more code and conditional codes or maybe I didn't find a good way to use it.
> The PR you mentioned didn't say any word about a standard, so I don't know what standard to follow. Besides, there exists a `useEtherPrice` in roadmap. Therefore I think it's fine to put it in core but maybe later there will be another package like `@usedapp/price` providing price hooks from Chainlink, CoinGecko, or even Binance, FTX.

The complication I referred to was the same thing you mentioned, using multiple calls which I think only one is required.
The [answer] should return a BigNumber according to the ABI, while I'm not sure if it would be formatted to the correct decimals, you still don't have to use parseFloat for some known reasons...

When I get on screen, might make some suggestions... 

Appreciate your contribution...

My words above shouldn't be seen as a "tackle", more like a suggestion... 

Never meant to undermine your contributions.

flyinglimao:
> When I get on screen, might make some suggestions...
> 
> Appreciate your contribution...
> 
> My words above shouldn't be seen as a "tackle", more like a suggestion...
> 
> Never meant to undermine your contributions.

Thank you for commenting and sorry for that I'm not a native English speaker therefore maybe there are some problem in my words and maybe I misunderstood / will misunderstand.

The `latestAnswer` is a BigNumber and should be divide with 10^(decimals), `decimals` is a `uint8` in ABI so it will be a number. Currently `eth-usd` return `284114061122` as `latestAnswer` and `8` as `deciamls`, so the price is `2841.14061122`.
Because BigNumber handle numbers in Integer domain, therefore I separate 2 parts, integral and fractional to get `2841` and `14061122`. And I want to provide it as number so used a `parseFloat`.
Although it's possible to cast `latestAnser` to number and divide directly as the range is seldom out of range of number in JavaScript, for robustness of code I prefer not to do so.

noobshow:
Returning amount in string is still more reliable than with number(either float or double) in Javascript and many other languages... floats have a safe maximum number in Javascript

packages/core/src/hooks/useChainlinkPriceFeed.ts
```diff
@@ -0,0 +1,32 @@
...
+    ) ?? []
+  const [answer] =
+    useContractCall(
+      address && {
+        abi: ChainlinkPriceFeedInterface,
+        address,
+        method: 'latestAnswer',
+        args: [],
+      }
+    ) ?? []
```
marekkirejczyk, line 24: Merge into one `useContractCalls`

packages/core/src/hooks/useChainlinkPriceFeed.ts
```diff
@@ -0,0 +1,32 @@
...
+  if (!decimals || !answer) return
+
+  const divider = BigNumber.from(10).pow(decimals)
+  const integral = answer.div(divider)
+  const fractional = answer.mod(divider)
+
+  return parseFloat(integral + '.' + fractional)
```
marekkirejczyk, line 31: It is better to return FixedNumber, i.e.
```ts
return FixedNumber.fromValue(answer, decimals)
```

marekkirejczyk:
Hi @flyinglimao,

Great work again! 🥇 
I added comments in the code.
I think it is a good idea to extract it a separate package like as #240 @noobshow proposed.
Looking forward to the changes and merge.

noobshow:
> > Hi @flyinglimao,
> > Great work again! 🥇
> > I added comments in the code.
> > I think it is a good idea to extract it a separate package like as #240 @noobshow proposed.
> > Looking forward to the changes and merge.
> 
> would you think the `@usedapp/price` is better than `@usedapp/chainlink` and `@usedapp/coingecko`? In my opinion, `@usedapp/price` is more common if there will be other price providers added to lib.

@usedapp/price would still require other price package to have a kind of prefix or suffix like @usedapp/price-chainlink | @usedapp/price-coingecko

But that would require a standard interface or not.

marekkirejczyk:
> would you think the `@usedapp/price` is better than `@usedapp/chainlink` and `@usedapp/coingecko`? In my opinion, `@usedapp/price` is more common if there will be other price providers added to lib.

The reasons to extract separate packages:

1. Isolate external dependencies. We don't want core to become a place with 1000s of dependencies, querying 100s of external servers.

2. Provide choice. The price can be obtained in many different ways (e.g. chainlink or coingecko). Some people prefer one place some other. So we want them to be able to choose, while brining minimal overhead with it. 

3. Mitigate inconsistencies. Each is a bit different and they are not consistent or compatible with each other. Let's not ask user to add a package with similar functions that get different arguments.

4. Additionally chainlink (or any other) might have other features in the future that we can add to the package (while keeping the same list od dependencies).

Therefore I believe the best way to split is around specific providers (coingecko, chainlink, coinbase), rather than functionality (price)

marekkirejczyk:
LGTM, need a rebase and it is ready to merge.

@Szymx95 @nezouse Any feedback on your side?

packages/chainlink/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@usedapp/chainlink",
+  "version": "0.3.21",
```
marekkirejczyk, line 3: We can actually number packages independently, so we can version this 1.0

PR #242, opened 2021-05-26
Introduce useUpdateConfig

Szymx95:
This PR will add `useSetHook` that returns `setConfig` function.

Set config accepts object as parameter that will be used to update config.

### DIFF
--- a/packages/core/src/providers/config/context.ts
+++ b/packages/core/src/providers/config/context.ts
@@ -1,9 +1,18 @@
 import { createContext, useContext } from 'react'
-import { FullConfig } from '../../model/config/Config'
+import { FullConfig, Config } from '../../model/config/Config'
 import { DEFAULT_CONFIG } from '../../model/config/default'
 
-export const ConfigContext = createContext<FullConfig>(DEFAULT_CONFIG)
+export const ConfigContext = createContext<{ config: FullConfig; setConfig: (config: Config) => void }>({
+  config: DEFAULT_CONFIG,
+  setConfig: () => undefined,
+})
 
 export function useConfig() {
-  return useContext(ConfigContext)
+  const { config } = useContext(ConfigContext)
+  return config
+}
+
+export function useSetConfig() {
+  const { setConfig } = useContext(ConfigContext)
+  return setConfig
 }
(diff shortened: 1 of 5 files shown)


packages/core/src/providers/config/context.ts
```diff
@@ -1,9 +1,18 @@
...
 
 export function useConfig() {
-  return useContext(ConfigContext)
+  const { config } = useContext(ConfigContext)
+  return config
+}
+
+export function useSetConfig() {
+  const { setConfig } = useContext(ConfigContext)
+  return setConfig
```
marekkirejczyk, line 17: Idea: Why not a single hook? ie
```ts
const {setConfig, ...config} = useConfig() 
```

or
```ts
const config = useConfig() 
config.setConfig()
```

or 
```ts
const {setConfig} = useConfig() 
```

Also:
`setConfig` or `updateConfig`?

marekkirejczyk (approved):

noobshow:
While this pull request is approved and merged, I doubt @Szymx95 understood the suggestion posted https://github.com/EthWorks/useDApp/pull/242#discussion_r639794078 that @marekkirejczyk was proposing the `setConfig` or `updateConfig` to be destructured from the same useConfig and make it look like this
`const {config, setConfig} = useConfig()`
Instead, he added a new hook to support useConfig job... I guess that's like duplicating or not a good coding practice...

Hopefully, it can be fixed as this seems to be a minor version and many interfaces are bound to change.

marekkirejczyk:
> While this pull request is approved and merged, I doubt @Szymx95 understood the suggestion posted [#242 (comment)](https://github.com/EthWorks/useDApp/pull/242#discussion_r639794078) that @marekkirejczyk was proposing the `setConfig` or `updateConfig` to be destructured from the same useConfig and make it look like this
> `const {config, setConfig} = useConfig()`

We went back and forth on this one with @Szymx95. Finally decided to be two separate hooks. The rationale is that one can use `config` quite often as a normal user. While `updateConfig` is reserved mostly for under-the-hood framework features. Also it feels right that useConfig returns just plain object without functions.

noobshow:
I think that's something to agree with...

PR #245, opened 2021-05-27
Localhost deploy multicall

OsamaElHariri:
**Deploy a Multicall contract if the user is working on a local Blockchain and a Multicall address is not provided.**

When a contract is deployed, the address is printed to the console and can be added to the configs.
The address is also added to the configs using `updateConfig`, so it is possible for the user to continue working without changing the configs

image

image

---
PS: This PR is a submission to 0xHackathon for the "[BEST NEW useDApp feature](https://gitcoin.co/issue/EthWorks/useDApp/181/100025575)" track


packages/core/src/providers/DAppProvider.tsx
```diff
@@ -28,15 +29,38 @@ interface WithConfigProps {
 }
 
 function DAppProviderWithConfig({ children }: WithConfigProps) {
+  const [localMulticallAddresses, setLocalMulticallAddresses] = useState({})
   const { multicallAddresses, readOnlyChainId, readOnlyUrls } = useConfig()
-  const multicallAddressesMerged = { ...MULTICALL_ADDRESSES, ...multicallAddresses }
+  const multicallAddressesMerged = { ...MULTICALL_ADDRESSES, ...multicallAddresses, ...localMulticallAddresses }
+
+  const setupMulticall = (chainId: ChainId) => {
```
Szymx95, line 36: Hi,

Great work,

May i suggest wrapping this function in 

```ts
const setupMulticall = useCallback((chainId: ChainId) => {
...
},[])
```

So that it won't be created at every render.

Szymx95:
Hello @OsamaElHariri ,

Great work with this feature but i'm not totally sure that deploying multicall contract should be a job of core package.
Personally i think that we should create a new package that runs local test environment that automatically deploys multicall but I think we should discuss this with @marekkirejczyk. 

In the meantime i think `const {deploy, contractAddress, state} = useDeployContract(ABI)` would be a great addition to core api.

packages/core/src/helpers/contract.ts
```diff
@@ -0,0 +1,33 @@
...
+        return deployMulticall(readOnlyUrls[chainId])
+    } else {
+        return
+    }
+}
+
+export async function deployMulticall(url: string): Promise<ContractDeployment> {
+    const provider = new ethers.providers.JsonRpcProvider(url)
+    const accounts = await provider.listAccounts()
+    const account = accounts[0]
```
marekkirejczyk, line 22:
```ts
function getDeploymentWallet(provider) {
  const [account] = await provider.listAccounts()
  return account
}
```

packages/core/src/helpers/contract.ts
```diff
@@ -0,0 +1,33 @@
...
+import { ChainId } from '../constants'
+import { NodeUrls } from '../model/config/Config'
+import multicallABI from '../constants/abi/MultiCall.json'
+import { isLocalChain } from '../helpers'
+
+interface ContractDeployment {
+    contractAddress: string
+}
+
+export async function deployLocalMulticall(chainId: ChainId, readOnlyUrls: NodeUrls | undefined): Promise<ContractDeployment | undefined> {
```
marekkirejczyk, line 11: deployMulticallIfLocal

packages/core/src/helpers/contract.ts
```diff
@@ -0,0 +1,33 @@
...
+    const accounts = await provider.listAccounts()
+    const account = accounts[0]
+
+    const txData = [{
+        "from": account,
+        "gas": "0x64b94",
+        "data": `0x${multicallABI.bytecode}`
+    }]
+
+    const transactionHash = await provider.send('eth_sendTransaction', txData)
```
marekkirejczyk, line 30: You can try this:
```ts
await provider.sendTransaction({
  from: account,
  data: `0x${multicallABI.bytecode}`
})
```
or even this:
```ts
const factory = new ContractFactory( multicallABI.abi , multicallABI.bytocde , account)
await factory. deploy()
```

packages/core/src/providers/DAppProvider.tsx
```diff
@@ -28,15 +29,38 @@ interface WithConfigProps {
...
+
+    deployLocalMulticall(chainId, readOnlyUrls)
+      .then(txContractDeploy => {
+        if (txContractDeploy === undefined) throw TypeError('Failed to deploy Multicall')
+        console.log(`Add the Multicall Address to the configs:
+{
+  multicallAddresses: {
+    [ChainId.Localhost]: "${txContractDeploy.contractAddress}"
+  },
+}`
```
marekkirejczyk, line 47: const {updateConfig} = useUpdateConfig()

packages/core/src/providers/DAppProvider.tsx
```diff
@@ -28,15 +29,38 @@ interface WithConfigProps {
 }
 
 function DAppProviderWithConfig({ children }: WithConfigProps) {
+  const [localMulticallAddresses, setLocalMulticallAddresses] = useState({})
   const { multicallAddresses, readOnlyChainId, readOnlyUrls } = useConfig()
-  const multicallAddressesMerged = { ...MULTICALL_ADDRESSES, ...multicallAddresses }
+  const multicallAddressesMerged = { ...MULTICALL_ADDRESSES, ...multicallAddresses, ...localMulticallAddresses }
```
marekkirejczyk, line 34: Let's introduce new component `<LocalMulitcall> </Multicall>`

marekkirejczyk:
Idea: Write a short information in the guide?

marekkirejczyk:
> Great work with this feature but i'm not totally sure that deploying multicall contract should be a job of core package.
> Personally i think that we should create a new package that runs local test environment that automatically deploys multicall but I think we should discuss this with @marekkirejczyk.
I proposed to extract a separate LocalMulticall component to isolate concerns. It will only do sth if on a local node.

> In the meantime i think `const {deploy, contractAddress, state} = useDeployContract(ABI)` would be a great addition to core api.
I would not include it in the scope of this task. We can add to roadmap.

OsamaElHariri:
@marekkirejczyk @Szymx95 I updated the PR. It now uses a `LocalMulticall` provider that deploys Multicall on the local chain before rendering the child components.

There is one issue I faced that I'd like your opinion on. When I deploy the contract and update the configs, the `blockNumber` in [ChainStateProvider](https://github.com/EthWorks/useDApp/blob/master/packages/core/src/providers/chainState/provider.tsx#L18) does not update for a while. This causes the invocaction of [multicall](https://github.com/EthWorks/useDApp/blob/master/packages/core/src/providers/chainState/provider.tsx#L47) to fail a couple of times until the block number is up to date.

To get around this, I added a check in the `LocalMulticall` provider so that it not only waits for the contract to be deployed, but also for the block number to update. Does this approach make sense? Or is there a way to force the block number to update?

packages/core/src/providers/LocalMulticallProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+export function LocalMulticallProvider({ children }: LocalMulticallProps) {
+  const updateConfig = useUpdateConfig()
+  const { multicallAddresses } = useConfig()
+  const { library, chainId } = useEthers()
+  const [isDeployingMulticall, setIsDeployingMulticall] = useState(false)
+  const [multicallBlockNumber, setMulticallBlockNumber] = useState<number>()
+  const blockNumber = useBlockNumber()
+
+  const multicallAddress =
+    chainId !== undefined && multicallAddresses !== undefined ? multicallAddresses[chainId] : undefined
```
Szymx95, line 22: Equivalent to
```ts
const multicallAddress = chainId && multicallAddresses[chainId]
```

packages/core/src/providers/LocalMulticallProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+        updateConfig({
+          multicallAddresses: {
+            [chainId]: contractAddress,
+          },
+        })
+        setMulticallBlockNumber(receipt.blockNumber)
+      }
+    }
+    deployContractIfLocal()
```
Szymx95, line 43:
```ts

const deployMulticall = async () => {
        const signer = library?.getSigner()
        if(!signer) {
            setIsDeployingMulticall(false)
            return
        }
        const { contractAddress, receipt } = await deployContract(multicallABI, signer)
        console.log(`Deploying Multicall with contract address "${contractAddress}"`)
        updateConfig({
          multicallAddresses: {
            [chainId]: contractAddress,
          },
        })
        setMulticallBlockNumber(receipt.blockNumber)
        setIsDeployingMulticall(false)
    }

if (isLocalChain(chainId) && !multicallAddress) {
  setIsDeployingMulticall(true)
  deployMulticall()
}

```

packages/core/src/providers/LocalMulticallProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const { library, chainId } = useEthers()
+  const [isDeployingMulticall, setIsDeployingMulticall] = useState(false)
+  const [multicallBlockNumber, setMulticallBlockNumber] = useState<number>()
+  const blockNumber = useBlockNumber()
+
+  const multicallAddress =
+    chainId !== undefined && multicallAddresses !== undefined ? multicallAddresses[chainId] : undefined
+
+  useEffect(() => {
+    if (isDeployingMulticall || multicallAddress !== undefined || library === undefined || chainId === undefined) {
```
Szymx95, line 25:
```ts
if (isDeployingMulticall || multicallAddress || !library || !chainId) {
```

packages/core/src/providers/LocalMulticallProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+          },
+        })
+        setMulticallBlockNumber(receipt.blockNumber)
+      }
+    }
+    deployContractIfLocal()
+  }, [library, chainId])
+
+  const awaitingMulticallBlock =
+    multicallBlockNumber !== undefined && blockNumber !== undefined && blockNumber < multicallBlockNumber
```
Szymx95, line 47:
```ts
multicallBlockNumber && blockNumber && blockNumber < multicallBlockNumber
```

packages/core/src/providers/LocalMulticallProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  if (chainId !== undefined && isLocalChain(chainId) && (!multicallAddress || awaitingMulticallBlock)) {
```
Szymx95, line 49:
```ts
if(chainId && ...)
```

docs/source/core.rst
```diff
@@ -49,6 +49,16 @@ Stores configurations and makes them available via `useConfig`_ hook.
 =====================
 
 
+<LocalMulticallProvider>
+=====================
+
+Ensures that a Multicall contract address is available when developing on a local chain. 
+A Multicall contract will be deployed when an address is not defined in the `Config`_.
```
Szymx95, line 56: A Multicall contract will be deployed when a multicall address on a local chainID is not defined in the `Config`_.

Szymx95:
Hi,

nice work so far,

looks good, i think this approach is fine since it's only for the local chain. I've written a few comments with suggestions.

OsamaElHariri:
Thanks for the comments :+1: . I fixed the `IsDeployingMulticall` code as per your comments, and cleaned up the `undefined` checks in this file

packages/core/src/helpers/contract.ts
```diff
@@ -0,0 +1,24 @@
...
+  receipt: ethers.providers.TransactionReceipt
+}
+
+export async function deployContract(
+  contractAbi: ContractAbi,
+  signer: ethers.providers.JsonRpcSigner
+): Promise<ContractDeployment> {
+  const factory = new ContractFactory(contractAbi.abi, contractAbi.bytecode, signer)
+  const contract = await factory.deploy()
+  const deployedContract = await contract.deployTransaction.wait()
```
marekkirejczyk, line 19: Or simply:
```ts
const receipt = await contract.deployTransaction.wait()
```

packages/core/src/helpers/contract.ts
```diff
@@ -0,0 +1,24 @@
...
+  return {
+    contractAddress: deployedContract.contractAddress,
+    receipt: deployedContract,
+  }
+}
```
marekkirejczyk, line 24: Simpler version:
```ts
export async function deployContract(
  contractAbi: ContractAbi,
  signer: providers.JsonRpcSigner
): Promise<TransactionReceipt> {
  const factory = new ContractFactory(contractAbi.abi, contractAbi.bytecode, signer)
  const contract = await factory.deploy()
  return await contract.deployTransaction.wait()
}
```

packages/core/src/providers/LocalMulticallProvider.tsx
```diff
@@ -0,0 +1,61 @@
...
+  children: ReactNode
+}
+
+export function LocalMulticallProvider({ children }: LocalMulticallProps) {
+  const updateConfig = useUpdateConfig()
+  const { multicallAddresses } = useConfig()
+  const { library, chainId } = useEthers()
+  const [isDeployingMulticall, setIsDeployingMulticall] = useState(false)
+  const [multicallBlockNumber, setMulticallBlockNumber] = useState<number>()
+  const blockNumber = useBlockNumber()
```
marekkirejczyk, line 19: I think we could simplify this method by introducing state enum:

```ts
enum LocalMulticallState {
  Unknown,
  NonLocal,
  LocalNotDeployed,
  Deploying,
  Deployed, 
  Error
}
...
function LocalMulticallProvider(...) {
... 
  useEffect(() => {
    if (!chainId) {
      setState(Unknown)
    } else if (isLocalChain(chainId)) {
      if (multicallAddresses[chainId]) {
        setState(Deployed)
      } else {
        setState(Deploing)
        const { contractAddress } = await deployContract(multicallABI, signer)
        //Also error handling here
        updateConfig({multicallAddresses: { [chainId]: contractAddress} })        
        setState(Deployed)
      }      
    } else {
      setState(NonLocal)
    } 
  }, [library, chainId])
...
  if ([NonLocal, Deployed].includes(state)) {
      return <>{children}</>
  } else if (state === Deploying) {
      return <div>Deploying multicall...</div>
  } else 
   ...
  
}
```

marekkirejczyk:
Hi @OsamaElHariri,

Sorry for slow review. I was away on Holidays. I looked at the latest version of code. Looks very good. 👌 
I proposed to refactoring and I think we should be ready to merge.
Let me know if you will find time or if you prefer that we do it.

Marek

OsamaElHariri:
Hey @marekkirejczyk , I went over the refactoring suggestions and updated the PR. I used the enum, and though there's an Error state, I don't have this component handle the error. Should this component handle an error when it comes to a failed multicall deploy by returning a component that just reads "Could not deploy multicall, deploy it manually and the contract address to the Configs" (or something similar)?

Also, while revisiting this PR I realized something. This statement `blockNumber < multicallBlockNumber` sometimes takes a good 10 - 15 sec to resolve, so the user is just staring at `Deploying multicall...` for a while, which makes it feel like it's stuck. I'm wondering if adding a loader here might be worth it?

packages/core/src/providers/LocalMulticallProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+      setLocalMulticallState(LocalMulticallState.Error)
+      return
+    }
+
+    setLocalMulticallState(LocalMulticallState.Deploying)
+
+    const deployMulticall = async () => {
+      try {
+        const { contractAddress, blockNumber } = await deployContract(multicallABI, signer)
+        console.log(`Deploying Multicall with contract address "${contractAddress}"`)
```
marekkirejczyk, line 60: I would avoid using console.log as much as possible to keep console clear.

marekkirejczyk:
> Hey @marekkirejczyk , I went over the refactoring suggestions and updated the PR. I used the enum, and though there's an Error state, I don't have this component handle the error. 
Just showing "Error deploying multicall contract" is fine.

> Also, while revisiting this PR I realized something. This statement `blockNumber < multicallBlockNumber` sometimes takes a good 10 - 15 sec to resolve, so the user is just staring at `Deploying multicall...` for a while, which makes it feel like it's stuck. I'm wondering if adding a loader here might be worth it?

It should not be required to use that comparison at all.

The sequence of events is:

1. useEffect setState to "Deploying"
2. contract is deployed
3. useEffect setState to "Deployed"
4. From now on we can render components (no need to wait for anything)

packages/core/src/providers/LocalMulticallProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+    if (localMulticallState === LocalMulticallState.Deploying) {
+      return
+    }
+
+    const signer = library.getSigner()
+    if (!signer) {
+      setLocalMulticallState(LocalMulticallState.Error)
+      return
+    }
```
marekkirejczyk, line 53: I think it would be nicer style to have `if ... else if` rather than `return` statements, i.e.
```ts
    if (!library || !chainId) {
      setLocalMulticallState(LocalMulticallState.Unknown)
    } else if (!isLocalChain(chainId)) {
      setLocalMulticallState(LocalMulticallState.NonLocal)
    } else if (multicallAddresses && multicallAddresses[chainId]) {
      setLocalMulticallState(LocalMulticallState.Deployed)
    } else
    ...
```

OsamaElHariri:
Updated! Refactored and removed the console log.

The blockNumber check is there because these error pop up when it's not, see image:

image


When I narrowed it down, I found that this error will continue to be thrown until the blockNumber returned from the `useBlockNumber()` hook equals the blockNumber in the contract deployment receipt. The error is thrown in the [providers/chainState/provider.tsx](https://github.com/EthWorks/useDApp/blob/b9304cbc951a50621aeec7b9be541ea0a10da406/packages/core/src/providers/chainState/provider.tsx#L33)

marekkirejczyk:
> When I narrowed it down, I found that this error will continue to be thrown until the blockNumber returned from the `useBlockNumber()` hook equals the blockNumber in the contract deployment receipt. The error is thrown in the [providers/chainState/provider.tsx]

That is strange, I'll investigate tomorrow.

packages/core/src/providers/LocalMulticallProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+      }
+      deployMulticall()
+    }
+  }, [library, chainId])
+
+  const awaitingMulticallBlock = multicallBlockNumber && blockNumber && blockNumber < multicallBlockNumber
+
+  if (
+    localMulticallState === LocalMulticallState.Deploying ||
+    (localMulticallState === LocalMulticallState.Deployed && awaitingMulticallBlock)
```
marekkirejczyk, line 63:
```suggestion
    (localMulticallState === LocalMulticallState.Deployed)
```

packages/core/src/providers/LocalMulticallProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+          setLocalMulticallState(LocalMulticallState.Deployed)
+        } catch {
+          setLocalMulticallState(LocalMulticallState.Error)
+        }
+      }
+      deployMulticall()
+    }
+  }, [library, chainId])
+
+  const awaitingMulticallBlock = multicallBlockNumber && blockNumber && blockNumber < multicallBlockNumber
```
marekkirejczyk, line 59:
```suggestion
```

marekkirejczyk:
> > When I narrowed it down, I found that this error will continue to be thrown until the blockNumber returned from the `useBlockNumber()` hook equals the blockNumber in the contract deployment receipt. The error is thrown in the [providers/chainState/provider.tsx]
> 
> That is strange, I'll investigate tomorrow.

I removed `awaitingMulticallBlock` and tested on Ganache and it works for me (see suggestions I added to the code).
Will it work for you? Are you testing on Ganache or sth else?

packages/core/src/providers/LocalMulticallProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+  Deployed,
+  Error,
+}
+
+export function LocalMulticallProvider({ children }: LocalMulticallProps) {
+  const updateConfig = useUpdateConfig()
+  const { multicallAddresses } = useConfig()
+  const { library, chainId } = useEthers()
+  const [localMulticallState, setLocalMulticallState] = useState(LocalMulticallState.Unknown)
+  const [multicallBlockNumber, setMulticallBlockNumber] = useState<number>()
```
marekkirejczyk, line 26:
```suggestion
```

OsamaElHariri:
I tried the suggestions but the errors still occur for me.

I'm using Ganache, and this also happens when I test it on HardHat. The OS I'm using is OpenSuse Leap 15.2, I downloaded the Ganache AppImage from their [web page](https://www.trufflesuite.com/ganache). I tried setting things up with Node version 16, version 14, and version 12 and all result in the same behaviour. I'm not sure what the difference between my setup and yours is :thinking: 

### This is what happens with me: when the page loads, I get that error every now and then:
image

---
### Then after a few seconds, I stop getting the error and all the contract infos are displayed:
image

marekkirejczyk:
I was able to reproduce - not sure where it is coming from yet.

marekkirejczyk (approved):

trozler:
@marekkirejczyk Hi quick question,
In order for useDapp to auto-deploy the multicall contract, is this config sufficient:
i.e. Do I need to set `readOnlyChainId`?

Also in general should you be setting `readOnlyChainId `?

```
const config = {
  readOnlyUrls: {
    [ChainId.Mainnet]: "https://mainnet.infura.io/v3/dbec3d12a1d64dcf8fb6c6f4536be321",
    [ChainId.Hardhat]: "http://localhost:8545",
  },
  multicallAddresses: {
    ...MULTICALL_ADDRESSES,
  },
  supportedChains: [
    ChainId.Mainnet,
    ChainId.Goerli,
    ChainId.Kovan,
    ChainId.Rinkeby,
    ChainId.Ropsten,
    ChainId.xDai,
    ChainId.Hardhat,
  ],
};
```

Brodan:
Is there some way to configure useDapp so that the multicall contract doesnt have to deploy every single time i refresh my page? Also it seems that I can't configure anything in my `HardhatUserConfig` so either the [documentation](https://usedapp.readthedocs.io/en/latest/core.html?highlight=multicallAddresses#config) is out of date or I am extremely confused. 

can someone ELI5 why i need to deal with this multicall stuff anywhere? I'm just trying to point my web app/metamask to my locally running eth node (running via hardhat). im pretty confused

sephynox:
> Is there some way to configure useDapp so that the multicall contract doesnt have to deploy every single time i refresh my page? Also it seems that I can't configure anything named `multicallAddresses` in my `HardhatUserConfig` so either the [documentation](https://usedapp.readthedocs.io/en/latest/core.html?highlight=multicallAddresses#config) is out of date or I am extremely confused.
> 
> can someone ELI5 why i need to deal with this multicall stuff anywhere? I'm just trying to point my web app/metamask to my locally running eth node (running via hardhat). im pretty confused

@Brodan I had this issue until updating to @usedapp/core:^0.10.0 which resolved it for me. Check your local geth terminal for the address your contracts are deploying to and use that for the `multicallAddresses` config. 

The config is for the DAppProvider and not hardhat. Here is a quick example: https://codesandbox.io/s/unruffled-newton-jpww2?file=/src/index.tsx

okorieebube:
> > Is there some way to configure useDapp so that the multicall contract doesnt have to deploy every single time i refresh my page? Also it seems that I can't configure anything named `multicallAddresses` in my `HardhatUserConfig` so either the [documentation](https://usedapp.readthedocs.io/en/latest/core.html?highlight=multicallAddresses#config) is out of date or I am extremely confused.
> > can someone ELI5 why i need to deal with this multicall stuff anywhere? I'm just trying to point my web app/metamask to my locally running eth node (running via hardhat). im pretty confused
> 
> @Brodan I had this issue until updating to @usedapp/core:^0.10.0 which resolved it for me. Check your local geth terminal for the address your contracts are deploying to and use that for the `multicallAddresses` config.
> 
> The config is for the DAppProvider and not hardhat. Here is a quick example: https://codesandbox.io/s/unruffled-newton-jpww2?file=/src/index.tsx

Heyy @sephynox , so I'm deploying three contracts, does that mean i'll supply all of them as my multicallAddresses. If yes, can you show me a code sample? Thanks

sephynox:
> > > Is there some way to configure useDapp so that the multicall contract doesnt have to deploy every single time i refresh my page? Also it seems that I can't configure anything named `multicallAddresses` in my `HardhatUserConfig` so either the [documentation](https://usedapp.readthedocs.io/en/latest/core.html?highlight=multicallAddresses#config) is out of date or I am extremely confused.
> > > can someone ELI5 why i need to deal with this multicall stuff anywhere? I'm just trying to point my web app/metamask to my locally running eth node (running via hardhat). im pretty confused
> > 
> > 
> > @Brodan I had this issue until updating to @usedapp/core:^0.10.0 which resolved it for me. Check your local geth terminal for the address your contracts are deploying to and use that for the `multicallAddresses` config.
> > The config is for the DAppProvider and not hardhat. Here is a quick example: https://codesandbox.io/s/unruffled-newton-jpww2?file=/src/index.tsx
> 
> Heyy @sephynox , so I'm deploying three contracts, does that mean i'll supply all of them as my multicallAddresses. If yes, can you show me a code sample? Thanks

Hello @okorieebube! The muilticall address could use with much better documentation. You can read more about multicalls here: https://github.com/joshstevens19/ethereum-multicall#readme but overall, it is used to reduce the number of JSON RPC requests an app makes. You can also checkout https://githubhelp.com/EthWorks/useDApp which discusses the proposal/implementation of multicall in useDApp.

However, for your purposes, no, you do not need to have separate entries for each contract. On one of my projects, I have "0x44691b39d1a75dc4e0a0346cbb15e310e6ed1e86" as the multicall address. I believe this address will also work if you run a forked development node off mainnet. Otherwise, let the multicall request transaction occur, you should then see something like this in your hardhat node terminal:

```eth_call
Contract call:       <UnrecognizedContract>
From:                0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
To:                  0xca8c8688914e0f7096c920146cd0ad85cd7ae8b9
```

Notice the "To" address. I used that as the multicall address when forking. However, you can also deploy the contract which is useDapp deploying a mulicall contract for you (Hence the error: "Error deploying multicall contract" if you reject). You should then see something like:

```
Contract deployment: <UnrecognizedContract>
Contract address:    0x976fcd02f7c4773dd89c309fbf55d5923b4c98a1
Transaction:         0xfec80142a354a8f52524c76fde32cf228d0458502c86d1b910047cde2997c467
From:                0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
Value:               0 ETH
Gas used:            408468 of 408468
Block #14169388:     0x80d60a2a5f580f2109883a393b53e00f9c939a8ac43125b55d51bc1a2c6b40c1
```
  
You can then just use the "Contract address" from this output as the multicall address. At least for me, either of them worked and I stopped getting constantly pestered to deploy a contract.

This, your multicallAddresses config for frontend would look like this:

```
...
    multicallAddresses: {
        [ChainId.Localhost]: "0x976fcd02f7c4773dd89c309fbf55d5923b4c98a1",
    }
...
```

okorieebube:
> 

Thanks soo much @sephynox I must say it makes much better sense now.
The documentation certainly needs more work, maybe I'll try adding to the documentation when I'm done with my project.
I really appreciate the help

PR #249, opened 2021-05-28
Introduce deep config update

Szymx95:


packages/core/src/providers/config/reducer.ts
```diff
@@ -1,5 +1,9 @@
 import { FullConfig, Config } from '../..'
 
 export function configReducer(state: FullConfig, action: Config): FullConfig {
-  return { ...state, ...action }
+  const readOnlyUrls = { ...state.readOnlyUrls, ...action.readOnlyUrls }
+  const multicallAddresses = { ...state.multicallAddresses, ...action.multicallAddresses }
+  const supportedChains = { ...state.supportedChains, ...action.supportedChains }
+  const notifications = { ...state.notifications, ...action.notifications }
+  return { ...state, ...action, readOnlyUrls, multicallAddresses, supportedChains, notifications }
```
marekkirejczyk, line 8: How about using [loadash merge](https://lodash.com/docs/4.17.15#merge).

You can the function it as a single dependecy.

marekkirejczyk (approved):

PR #257, opened 2021-06-01
💵 Introduce @usedapp/uniswap with useUniswapPrice hook

massun-onibakuchi:
Uniswap is not only a simple exchange for tokens, but also has the function of PriceOracle. This PR introduces @usedapp/uniswap and a hook
` useUniswapPrice(baseCurrency : string ,quateCurrency: string , overrides?:{factory:address,digits:number}) => FixedNumber | undefined`

This hook will return a peek price from the Uniswap V2 pool, calculated from the address of the given token. If there is no direct path or unconnected, it returns undefined. This hooks doesn't depend on Uniswap SDK. I think that in most cases, we just want to know the price, so we don't need to think about complicated paths.
uniswap determines the price from the pool reserve (e.g ETH/DAI, ETH-DAI price = the reserve of Dai / the reserve of WETH)
By overriding factory address, this hook can use sushiswap as a price feed.

 -  extract the specific changes
 -  linting and formatting passed


packages/testing/examples/useUniswapPrice.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  function sortAddress(tokenA: string, tokenB: string) {
+    return compareAddress(tokenA, tokenB) === -1 ? [tokenA, tokenB] : [tokenB, tokenA]
+  }
+
+  beforeEach(async () => {
+    tokenA = await deployMockToken(deployer)
+    tokenB = await deployMockToken(deployer)
+    ;({ factory, pair } = await deployUniswapV2Pair(deployer, tokenA, tokenB))
```
noobshow, line 42: There's a semicolon here, should it be there?

packages/testing/examples/useUniswapPrice.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+    const computedAddress = getCreate2Address(factory.address, salt, INIT_CODE_HASH)
+    expect(computedAddress).to.equal(pair.address)
+  })
+
+  it('get price', async () => {
+    const [token0Addr] = sortAddress(tokenA.address, tokenB.address)
+
+    // base/quate (e.g. ETH/DAI): price of baseToken in quateToken = quateTokenReserve / baseTokenReserve
+    const [numerator, denominator] = tokenA.address === token0Addr ? [ONE, RATIO] : [RATIO, ONE]
+    const price = numerator.mul(EXP_SCALE).div(denominator)
```
noobshow, line 64: You can use parseFixed() from @ethersproject since it seems to do the same thing you want?

packages/testing/examples/useUniswapPrice.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+    const { result, waitForCurrent } = await renderWeb3Hook(
+      () => useUniswapPrice(tokenA.address, tokenB.address, { factory: factory.address }),
+      {
+        mockProvider,
+      }
+    )
+    await waitForCurrent((val) => val !== undefined)
+    expect(result.error).to.be.undefined
+    expect(result.current?.toString()).to.eq(FixedNumber.fromValue(price, DIGITS).toString())
```
noobshow, line 74: ParseFixed() could work here too

noobshow: With or without using BigNumber, all seems well except for the failing test...

BigNumber seems to be preferable as a return value than FixedNumber since many hooks exported usually have it exposed instead and they tend to get returned directly from ethers.js contract calls numeric values

packages/testing/examples/useUniswapPrice.test.tsx
shota-masuda-onibakuchi: First, I think so too, but It seems to require a semicolon. If there's  no semicolon, compilation will fail.
``` 
This expression is not callable.
  Type 'Promise<Contract>' has no call signatures.

41     tokenB = await deployMockToken(deployer)
                      ~~~~~~~~~~~~~~~~~~~~~~~~~

  examples/useUniswapPrice.test.tsx:41:20
    41     tokenB = await deployMockToken(deployer)
                          ~~~~~~~~~~~~~~~~~~~~~~~~~
    Are you missing a semicolon?
```

MDN's js reference says about assignment without declaration
> The parentheses ( ... ) around the assignment statement are required when using object literal destructuring assignment without a declaration. ..
...
Your ( ... ) expression needs to be preceded by a semicolon or it may be used to execute a function on the previous line.

packages/testing/examples/useUniswapPrice.test.tsx
shota-masuda-onibakuchi: I understand what you are saying. So, should the hook return 10^x times the price because otherwise I don't think it will work if the price is fractional?
In other words, if the actual price is 0.2021, should the hook return a BigNumber of 20210000 (where x is 8)?

noobshow: I couldn't fork on my machine to test too, but here's my thought about `BigNumber` vs `FixedNumber`

Behind the scene : 

`FixedNumber.fromValue(price, DIGITS)` will convert the parameters into string with `formatFixed(value, decimals)`,
then the return value get parsed back into a `BigNumber `with `parseFixed(value, decimals)`.

Your implementation works too, this is just a detail anyways.

packages/uniswap/src/hooks/useUniswapPrice.ts
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * @param {string} baseCurrency address of base token (e.g ETH address)
+ * @param {string} quateCurrency address of quate token (e.g DAI address)
+ * @param {Object} overrides optional factory address (For Sushiswap, default Uniswap V2 Factory)
+ * @returns {FixedNumber | undefined} peek price
+ */
+export function useUniswapPrice(
+  baseCurrency: string,
+  quateCurrency: string,
+  overrides?: { factory?: string; digits?: number }
```
werg, line 22: Could you add the option to override the init code hash here as well? All the uniswap-v2 compatible AMMs (Sushi, pancake QS..) have different init hashes as well as different factory addresses.

pawelpolak2 (approved):

PR #261, opened 2021-06-14
Refactor coingecko

Szymx95:


packages/coingecko/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@usedapp/coingecko",
-  "version": "0.3.21",
+  "version": "0.1.0",
```
marekkirejczyk, line 3: Now that we merged 0.3.21, we have to keep it consistent.

marekkirejczyk: My bad

marekkirejczyk (approved):

PR #270, opened 2021-06-20
feat: add ENS lookup hook

quagliero:
useLookupAddress returns the registered ENS name for a connected wallet address (or undefined).

Screenshot 2021-06-20 at 19 37 42
Screenshot 2021-06-20 at 19 38 00


noobshow:
> useLookupAddress returns the registered ENS name for a connected wallet address (or undefined).
> 
> Screenshot 2021-06-20 at 19 37 42
> Screenshot 2021-06-20 at 19 38 00

Shouldn't it be something along the line of `useENSAddress `or `useENSName`?
Just a suggestion!

quagliero:
> Shouldn't it be something along the line of `useENSAddress `or `useENSName`?
> Just a suggestion!

@noobshow For the sake of consistency I named it after the provider method it wraps :) https://docs.ethers.io/v5/single-page/#/v5/api/providers/provider/-%23-Provider-lookupAddress

noobshow:
> > Shouldn't it be something along the line of `useENSAddress `or `useENSName`?
> > Just a suggestion!
> 
> @noobshow For the sake of consistency I named it after the provider method it wraps :) https://docs.ethers.io/v5/single-page/#/v5/api/providers/provider/-%23-Provider-lookupAddress

Oops, didn't see that...
Nevertheless, while that's a good way to name it, I only thought since this is meant to be a standalone function or library, it needs to communicate its uses in the hook naming...

Just my thought, I could look up a web address or a ethereum address... But, when it includes ENS, I would know for sure seeing the hook means it's looking up an ENS address

docs/source/core.rst
```diff
@@ -301,7 +301,7 @@ Returns connection state and functions that allow to manipulate the state.
 *Requires:* ``<ConfigProvider>``
 
 useGasPrice
-=========
+===========
```
marekkirejczyk, line 304: 👍

packages/example/src/components/account/AccountButton.tsx
```diff
@@ -2,12 +2,15 @@ import React, { useEffect, useState } from 'react'
 import { useEthers, shortenAddress } from '@usedapp/core'
 import { Button } from '../base/Button'
 import { Colors } from '../../global/styles'
+// @TODO replace once merged and published into @usedapp/core
```
marekkirejczyk, line 5: You can just use:
`import { useLookupAddress } from `@usedapp/core``

First you need to run:
'yarn build`
form main repo directory thou

marekkirejczyk:
Hi @quagliero,

Great work! ❤️ 
Need to fix TODO, and I will merge (I left the comment).

Few follow-up considerations:
- should we take ENS from current network or always from mainnet?
- can we make it work on localhost? (see #245, I guess deploy local ENS, it sounds like overkill thou) 
- should we recheck the name when new block arrives? (I guess not, perhaps optionally)

Marek

marekkirejczyk (approved):

wminshew:
based on local testing it seems like library is a new object every re-render & so re-triggers the api call every render -- is there something else that should be used as the dependency in this hook? or should we memoize library inside usedapp?

noobshow:
I think you can memoize the particular hooks used in your project, for
example memoize your calls to useContractCalls(and many more like that),
the problem will reduce or be solved depending on your project and
memoization method.

vm06007:
this only works for connected wallet, is there way to pass address as a parameter to lookup someone else ENS name?

noobshow:
> this only works for connected wallet, is there way to pass address as a parameter to lookup someone else ENS name?

You think a useHook would be a good idea for that?

vm06007:
perhaps `useLookupAddress(ENS_NAME)` would be good idea. Forced to use `provider.lookupAddress(ENS_NAME)` using ethers directly.

vm06007:
hopefully this can be improved so `useLookupAddress()` can have a parameter

PR #277, opened 2021-07-07
fix(devtools): Adding a check for window object

ChangoMan:
Adds a check to see if `window` is defined, which prevents an error on Next.js due to server side rendering.


ChangoMan:
@marekkirejczyk May I request a review on this please. Thanks for your time!

marekkirejczyk (approved):

PR #281, opened 2021-07-15
Memoize hook return values to prevent unnecessary re-renders

### LINKED ISSUE #280: Hooks re-rendering on any state update
Tested on both 0.3.24 and 0.4.1, any state update in a component will make any hook that uses `useChainCalls` re-render.

Sample, simply adding a state-changing input on the getting ready example:

```js
import React, { useEffect, useState } from "react";

import { formatEther } from "@ethersproject/units";
import { useEtherBalance, useEthers } from "@usedapp/core";

const App: React.FC = () => {
  const [input, setInput] = useState<string>("");
  const { activateBrowserWallet, account } = useEthers();
  const etherBalance = useEtherBalance(account);

  useEffect(() => {
    console.log("balance updated!");
  }, [etherBalance]);

  return (
    <div>
      <div>
        <button onClick={() => activateBrowserWallet()}>Connect</button>
        <input value={input} onChange={e => setInput(e.currentTarget.value)} />
      </div>
      {account && <p>Account: {account}</p>}
      {etherBalance && <p>Balance: {formatEther(etherBalance)}</p>}
    </div>
  );
};

export default App;
```

Whenever `input` is changed, `etherBalance` will also update and the console log will be triggered. This should be easily reproduceable on the latest build of `@usedapp/core`.

It's a major side effect and making development using useDApp very problematic.

sns-sem:
Fixes #280 causing certain hooks to update on any parent component re-render because return values were not being memoized in any way.

Passes all tests and confirmed it fixes the issues as far as I could see

### DIFF
--- a/.changeset/chilly-timers-chew.md
+++ b/.changeset/chilly-timers-chew.md
@@ -1,0 +1,5 @@
+---
+"@usedapp/core": patch
+---
+
+Memoize hook return values to prevent unnecessary re-renders
--- a/.changeset/nice-buttons-sit.md
+++ b/.changeset/nice-buttons-sit.md
@@ -1,0 +1,5 @@
+---
+"@usedapp/core": patch
+---
+
+Memoize hook return values to prevent unnecessary re-renders
--- a/.changeset/nice-buttons-sit.md
+++ b/.changeset/nice-buttons-sit.md
@@ -1,5 +1,0 @@
----
-"@usedapp/core": patch
----
-
-Memoize hook return values to prevent unnecessary re-renders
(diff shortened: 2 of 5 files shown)


marekkirejczyk:
@nezouse @Szymx95 Would you take a look at it?

nezouse:
Seems fine for me, actually wanted to do this as well. It's not really possible to correctly mark all hooks dependencies without this change.

Szymx95 (commented):
LGTM, but we need to think what can we do with those `JSON.stringify` in dependencies becouse this way seems like a hack, but it's fine for now

marekkirejczyk (approved):

PR #285, opened 2021-07-21
feat: better error messaging

quagliero:
Check for the existence of rpc error messages and use them, this prevents the mass "cannot estimate gas" messaging for errors that aren't related to gas.

Fixes https://github.com/EthWorks/useDApp/issues/260

PR #288, opened 2021-07-22
feat: better error messaging [previous version got overwritten]

quagliero:
Use the full errors from RPC client if error object exists


quagliero:
This addition got removed when the memoisation PR went on top of it.

https://github.com/EthWorks/useDApp/pull/285

marekkirejczyk (approved):

PR #300, opened 2021-08-12
feat: filter contract logs for events

lcamargof:
### Description of changes
Avoid `No Matching event` by only calling `parseLog` on the logs that correspond to the current contract called.

### Notes:
I'm not sure if this is the best solution for the problem, ideally we should return the logs unparsed but that can be found in the transaction itself

### DIFF
--- a/.changeset/eighty-worms-itch.md
+++ b/.changeset/eighty-worms-itch.md
@@ -1,0 +1,5 @@
+---
+'@usedapp/core': minor
+---
+
+Prevent crash when parsing transaction logs from contract call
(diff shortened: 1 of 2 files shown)


quagliero:
Just stumbled across this as having the same error. Nice one.

quagliero:
For anyone else arriving here, this error was introduced in the most recent relase (0.4.4), so if you fix to version `0.4.3` you're OK.

Szymx95 (commented):
Maybe we should pass the filtering function as argument to useContractFunction

```ts
useContractFunction(..., filteringFunction?: (e:any) => boolean)
```

So that developers can have more control and as default it can be left filtering address of contract

nezouse (approved):

lcamargof:
> Maybe we should pass the filtering function as argument to useContractFunction
> 
> ```ts
> useContractFunction(..., filteringFunction?: (e:any) => boolean)
> ```
> 
> So that developers can have more control and as default it can be left filtering address of contract

More than a filtering function I would say it's a parse function, where manually they read the logs and parse the event with the corresponding contract Abi, this could be an empty function to get the raw logs

PR #323, opened 2021-09-03
Add missing lodash-merge dependency

### LINKED ISSUE #321: Missing `lodash` dependency
## Expected behavior

I'm able to install `@usedapp/core` without any issues.

## Current behavior

Installing `@usedapp/core@0.4.7` throws a `Module not found: Can't resolve 'lodash'` error.

image

## Possible solution

Remove `lodash` dependency and use `Object.assign` instead of `merge`, which is only used [once](https://github.com/EthWorks/useDApp/blob/0011fe5e4428cb37f6efa870728e93e7cbff108f/packages/core/src/providers/config/reducer.ts#L2). Alternatively, include `lodash` in your `package.json` dependency, as it seems that just adding `@types/lodash` might not be enough.

## Steps to reproduce

1. `mkdir tmp && cd $_ && npm init -y && npm install @usedapp/core@0.4.7`
2.  See `lodash` missing in `node_modules`

## Context

I was trying to use `@usedapp/core` in a new project and not the usual scaffold I usually use, so for the first time I was using `latest`. Then my server showed the error. I couldn't help to notice that #249 added `lodash` in [here](https://github.com/EthWorks/useDApp/pull/249/files#diff-037b0ef349d759ea867ae708d180ef398f4b15147563bf8469b91df339f4cb02R5) but didn't add it on package.json. This was likely not picked up because it's such a common dependency that it's likely within another dependency. Oddly enough, though, this could had been picked up with a CI system or a simple GitHub action to build the project.

## Additional comments

Installing `lodash` in my project as a dependency solves the issue, but that seems like a patch rather than solving the actual problem. Also, we could just import `lodash.merge` instead of the whole thing. FWIW I went down versions as low as `0.4.1` and kept stumbling across this problem. Considering the constant updates and vulnerabilities that `lodash` keeps showing (or at least for how often I get spammed by Dependabot across my multiple projects), I would try to stir away from it in the future unless strickly necessary.

nezouse:
Fixes #321

### DIFF
--- a/.changeset/moody-cycles-think.md
+++ b/.changeset/moody-cycles-think.md
@@ -1,0 +1,5 @@
+---
+"@usedapp/core": patch
+---
+
+Add missing lodash-merge dependency
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -15,8 +15,9 @@
     "@ethereum-waffle/provider": "^3.2.2",
     "@types/chai": "^4.2.14",
     "@types/chai-as-promised": "^7.1.3",
-    "@types/mocha": "^8.2.0",
     "@types/lodash": "^4.14.170",
+    "@types/lodash.merge": "^4.6.6",
+    "@types/mocha": "^8.2.0",
     "@typescript-eslint/eslint-plugin": "^4.10.0",
     "@typescript-eslint/parser": "^4.10.0",
     "@web3-react/abstract-connector": "^6.0.7",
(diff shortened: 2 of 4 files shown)


jjperezaguinaga:
PR seems to be passing and everything seems working, anything externals like me can help to get this one merged?

marekkirejczyk (approved):

PR #327, opened 2021-09-09
🚀 More runs for lower gas costs

RetroDefi:
Mumbai polygon contract change to a compiled deployment with more runs in optimization to obtain lower gas costs


marekkirejczyk:
Hi @RetroDefi,

I don't think there is a contract under this address:
https://polygon-explorer-mumbai.chainstacklabs.com/address/0xD0Fd8713101E0d75dbc99C9FD6C10dE7451Aa12c/transactions

Also multicall contract only process read-only calls, so no gas is used.

PR #339, opened 2021-09-23
Add transactions path in local storage to config

vanruch:
We had a problem that transactions field in local storage has already been used and feel like this could be a problem for others as well. 
It's easier to override useDapps LS path than to perform a local storage migration


rzadp (approved):
I recommend adding unrelated cosmetic changes on separate PR in the future.

vanruch:
@rzadp Not sure what happened with these trimmed whitespaces, probably IDE removed them

Issue #349, opened 2021-10-05
no matching event when using useContractFunction with proxy


yehjxraymond:
When using the `useContractFunction` with a proxy contract, there may be events that cannot be parsed with the `contract.interface.parseLog` function in `@usedapp/core/src/hooks/useContractFunction.ts` since the event's signature is not found on the proxy contract.

## Code

The related code is:
```
export function useContractFunction(contract: Contract, functionName: string, options?: TransactionOptions) {
  const { library, chainId } = useEthers()
  const { promiseTransaction, state } = usePromiseTransaction(chainId, options)
  const [events, setEvents] = useState<LogDescription[] | undefined>(undefined)

  const send = useCallback(
    async (...args: any[]) => {
      const contractWithSigner = connectContractToSigner(contract, options, library)
      const receipt = await promiseTransaction(contractWithSigner[functionName](...args))
      if (receipt?.logs) {
        const events = receipt.logs
          .filter((log) => log.address === contract.address)
          .map((log) => contract.interface.parseLog(log))
        setEvents(events)
      }
    },
    [contract, functionName, options, library]
  )

  return { send, state, events }
}
```

## Suggestion

The library can perform an error handling on the `parseLog` to ignore events that cannot be parsed by the contract interface.
Happy to submit a PR to fix this. 

Cheers (=

PR #350, opened 2021-10-05
fix: event parsing for proxies on useContractFunction

yehjxraymond:
Proposed fix for issue #349


marekkirejczyk:
Lgtm ❤️

@Szymx95?

@yehjxraymond would you add test and changest?

Szymx95:
Lgtm

PR #362, opened 2021-10-13
🦞 Skip multicall if address is undefined

vanruch:
We had some nasty debugging when all useDapp calls stoped working after switching to Ropsten. Turns out, one of addresses was undefined for ropsten and all we got was Invalid ENS Name error in console. 

**Maybe we should also test if address has correct format to handle this better?**


marekkirejczyk:
LGTM, @Szymx95 @nezouse?

marekkirejczyk (approved):

PR #366, opened 2021-10-19
👻 Adds Fantom Testnet support

rafaeldalpra:
Adding support for Fantom Testnet and fixing a test case from the previous commit adding Fantom support.

I'm not sure how to handle/where to find the multicall address missing in my commit.

Any help/tips are appreciated!


tt-marek:
Continued in #414

PR #367, opened 2021-10-19
Add missing import to example in Getting Started

hasparus:


marekkirejczyk:
Good catch, merged from the other PR #361

PR #368, opened 2021-10-19
❄️ Add Avalanche support

werg:


packages/core/src/constants/chainId.ts
```diff
@@ -57,6 +59,7 @@ export const MULTICALL_ADDRESSES = {
   [ChainId.Harmony]: '0xFE4980f62D708c2A84D3929859Ea226340759320',
   [ChainId.Palm]: '0x99a73dfE34578348fb81BD078201C0BA84E9c840',
   [ChainId.Fantom]: '0xdc85396592f0F466224390771C861EE3957a3ff4',
+  [ChainId.Avalanche]: '0x0FB54156B496b5a040b51A71817aED9e2927912E',
```
tt-marek, line 62: Is that a real multicall contract address?
I am not an expert on Avalanche, but explorer seems not to show anything here:
https://explorer.avax.network/search?query=0x0FB54156B496b5a040b51A71817aED9e2927912E

werg: That is the multicall contract which I believe was deployed by the Pangolin team, but Avalanche recently made a deal with the etherscan people and completely replaced their blockchain explorer **apparently without migrating any of their contract verifications** (btw. the explorer you were searching on was for their EVM-incompatible subchain which is why it didn't show up as a contract, you want to instead use [Snowtrace](https;//snowtrace.io)).
Anyway -- I was unable to guess the right compiler version to verify this multicall contract in the explorer, however I was able to verify [this](https://snowtrace.io/address/0xddcbf776df3de60163066a5dddf2277cb445e0f3#code) MulticallV2 contract (I believe deployed by the [sushiswap](https://github.com/sushiswap/sushiswap-sdk/blob/canary/src/constants/addresses.ts#L340) team), with the code from the MakerDAO multicall repo, so I'll replace the address.

werg: OK I replaced it

tt-marek (approved):

PR #373, opened 2021-10-25
Make persistent connection optional

0xVenge:
Eventually it would be good to have a last-enabled provider persistence, but requiring the injected provider to persist makes implementation more difficult.


tt-marek:
LGTM @Szymx95 @nezouse 

@0xVenge Would you add documentation?

tt-marek:
Fixed with #385 (very similar implementation, but with true as default)

Issue #393, opened 2021-11-17
[refactor] define chain-specific config under modal/chain/


gasolin:
## Problem:

currently to define a chain require change multiple places:

- packages/core/src/constants/chainId.ts
 ChainId, CHAIN_NAMES, MULTICALL_ADDRESSES, TEST_CHAINS

- packages/core/src/model/config/default.ts
add supportedChains

- packages/core/src/helpers/chain.ts
update getExplorerAddressLink, getExplorerTransactionLink

- add test cases

and adding any new chain will conflict with the later PR

## Try:

We can define chain-specific config under `modal/chain/` with the following structure

```
modal/chain/
- ethereum.js
-bsc.js
```

and collect chain specific instances like

```js
export const Mainnet = {
    chainId: 1,
    chainName: 'Mainnet',
    isTestChain: false,
    isLocalChain: false,
    MULTICALL_ADDRESS: '0xeefba1e63905ef1d7acba5a8513c70307c1ce441',
    getExplorerAddressLink: (address: string) => `https://mainnet.etherscan.io/address/${address}`,
    getExplorerTransactionLink: (transactionHash: string) => `https://mainnet.etherscan.io/tx/${transactionHash}`,
}
```

currently, we can refactor parts by parts to keep the backward compatibility and ease the refactoring work. 

Phase 2:

Then to refactor `packages/core/src/helpers/chain.ts` to get results from the related chain object instead of from CHAIN_NAMES, TEST_CHAINS, LOCAL_CHAINS

and we can add `supportNetworks ` in `packages/core/src/model/config/Config.ts` to pass chain objects, and deprecate `multicallAddresses, supportChains` config props in the future version

Phase 3:

then we can define chains as separate packages like `@usedapp/polygon`, `@usedapp/bsc`, and use as 

```html
import ethereum from '@usedapp/ethereum'
import bsc from '@usedapp/bsc'

<DAppProvider config={{supportNetworks: [...ethereum, ...bsc]}} >
```

Issue #346, opened 2021-09-30
What's the standard way of adding chains?


Oni-giri:
UseDapp provides a limited set of supported chains. However I can't seem to be able to configure it to work with alternative chains: is it a design choice or do I miss the "right" way of doing it? 

Currently, the only way I see is to fork the repo and add manually the chain data in the `constants` folder.

Here is my (unfortunate) setup for arbitrum:

```javascript
const config = {
  readOnlyChainId: 42220,
  supportedChains: [42220],
  multicallAddresses: {
    42220: '0xCe129333458f037867C81DFbD70E4D57A3b1c5eb',
}, 
  readOnlyUrls: {
    42161: 'https://arb-mainnet.g.alchemy.com/v2/API-KEY',
  },
}
```

While the DApp can connect to Metamask, it is unable to load the provider it appears as "undefined".

If there's no good way to do it, I guess adding support to manage it in the config should solve a good amount of issues, past, present and future.

christianvari:
Agree with you, having the same problem developing a dapp on Cronos testnet

Oni-giri:
> @Oni-giri Yes I've found you can fake it to a degree using your own local version of `ChainId` that you extend, but it will ultimately fall over if the chain isn't in the core code. I suggest opening a PR like this one: https://github.com/EthWorks/useDApp/pull/336/files

Since the owner isn't replying, I forked the repo and uploaded it on NPM, so that I don't have to think about it any more.

Oni-giri:
@yuyao17 

You can either use https://www.npmjs.com/package/ethers-allstars-multicall, which is my fork of the package, or add those addresses, if you want to build it yourself:

const multicallAddresses = {
  1: '0xeefba1e63905ef1d7acba5a8513c70307c1ce441',
  3: '0xF24b01476a55d635118ca848fbc7Dab69d403be3',
  4: '0x42ad527de7d4e9d9d011ac45b31d8551f8fe9821',
  5: '0x77dca2c955b15e9de4dbbcf1246b4b85b651e50e',
  42: '0x2cc8688c5f75e365aaeeb4ea8d6a480405a48d2a',
  56: '0x1Ee38d535d541c55C9dae27B12edf090C608E6Fb',
  100: '0xb5b692a88bdfc81ca69dcb1d924f59f0413a602a',
  137: '0xc4f1501f337079077842343Ce02665D8960150B0',
  1337: '0x77dca2c955b15e9de4dbbcf1246b4b85b651e50e',
[...]

Issue #47, opened 2021-02-26
Replace enum ChainId with a set of constants. Replace ChainId with number


sz-piotr:
Using ChainId limits the possibilities of the users to interact with other chains. If I had a chain with id 777 I wouldn't be able to use the library.

marekkirejczyk:
btw list of chains, we could use to have a comprehensive support

https://chainid.network/chains.json

tarrencev:
In line with this suggestion, we should consider using chain agnostic identifiers following [CAIP standards](https://github.com/ChainAgnostic/CAIPs). These seem to be the standards with the most momentum behind them (wallet connect/metamask).

Benek7831:
Hi guys, 

As I understand, it is still not possible to use useDapp with chains other than the ones available through _ChainId_ ?
So fantom is available but not the testnet, and avalanche too, right ?

tt-marek:
Continued in #408

PR #394, opened 2021-11-17
🔗 Define chain specific configs

gasolin:
the first part of #393 , 
- refactor chain related config under modal/chain
- auto mapping getExplorerAddressLink(address)/getExplorerTransactionLink 

to support a new chain: dev could 
1. define a chain profile
2. import in packages/core/src/constants/chainId.ts
3. add tests

yarn build/yarn test passed


packages/core/src/constants/chainId.ts
```diff
@@ -20,11 +22,11 @@ export enum ChainId {
...
-  [ChainId.Ropsten]: 'Ropsten',
-  [ChainId.Kovan]: 'Kovan',
-  [ChainId.Rinkeby]: 'Rinkeby',
-  [ChainId.Goerli]: 'Goerli',
+  [Mainnet.chainId]: Mainnet.chainName,
+  [Ropsten.chainId]: Ropsten.chainName,
+  [Rinkeby.chainId]: Rinkeby.chainName,
+  [Goerli.chainId]: Goerli.chainName,
+  [Kovan.chainId]: Kovan.chainName,
   [ChainId.BSC]: 'BSC',
```
gasolin, line 30: in future phase we can use access chains object via index instead of define CHAIN_NAMES, MULTICALL_ADDRESSES...

packages/core/src/helpers/chain.ts
```diff
@@ -1,25 +1,18 @@
...
 
 export function getExplorerAddressLink(address: string, chainId: ChainId) {
   switch (chainId) {
-    case ChainId.Mainnet:
-    case ChainId.Ropsten:
-    case ChainId.Kovan:
-    case ChainId.Rinkeby:
-    case ChainId.Goerli:
-      return `https://${etherscanNetworkPrefix(chainId)}etherscan.io/address/${address}`
+    case Mainnet.chainId:
```
gasolin, line 6: can do `supportChains[chainId].getExplorerAddressLink(address)` once all migrated

gasolin:
@tt-marek please give some feedback

though the PR does not change or add new features but propose a new Chain definition format,
PR will get conflict with new chains support like #368

tt-marek:
Looks very promising, I'll add comments in the code.
Would be good to rebase on master and add more networks (there was a few PRs with new networks)

packages/core/src/constants/chainId.ts
```diff
@@ -20,58 +54,9 @@ export enum ChainId {
...
-  ChainId.Goerli,
-  ChainId.Mumbai,
-  ChainId.ThetaTestnet,
-  ChainId.Localhost,
-  ChainId.Hardhat,
-]
+export const MULTICALL_ADDRESSES: { [index: number]: string } = {}
+SUPPORT_NETWORKS.filter((network) => network.MULTICALL_ADDRESS).map(
+  (network) => (MULTICALL_ADDRESSES[network.chainId] = network.MULTICALL_ADDRESS)
+)
```
tt-marek, line 60: I would propose to introduce a function `getMulticallAddressFor` to be used instead of MULTICALL_ADDRESSES constant, what do you think?

tt-marek: And a helper function `getNetworkData(chainId)`

packages/core/src/model/chain/Chain.ts
```diff
@@ -0,0 +1,9 @@
+export type Chain = {
+  chainId: number
+  chainName: string
+  isTestChain: boolean
+  isLocalChain: boolean
+  MULTICALL_ADDRESS: string
+  getExplorerAddressLink: (address: string) => string
+  getExplorerTransactionLink: (address: string) => string
+}
```
tt-marek, line 9: 👍

packages/core/src/helpers/chain.ts
```diff
@@ -1,95 +1,26 @@
...
-    case ChainId.Avalanche:
-      return `https://snowtrace.io/tx/${transactionHash}`
-  }
+  const networks = SUPPORT_NETWORKS.filter((network) => network.chainId === chainId)
+  return networks[0]?.getExplorerTransactionLink(transactionHash) || ''
 }
 
 export function getChainName(chainId: ChainId) {
-  return CHAIN_NAMES[chainId]
+  const networks = SUPPORT_NETWORKS.filter((network) => network.chainId === chainId)
```
tt-marek, line 14: Use `find` instead of `filter` if you expect to find a single element

packages/core/src/helpers/chain.ts
```diff
@@ -1,95 +1,26 @@
...
+  return networks[0]?.chainName || ''
 }
 
-export function isTestChain(chainId: ChainId) {
-  return TEST_CHAINS.includes(chainId)
-}
+export const isTestChain = (chainId: ChainId) =>
```
tt-marek, line 18: It could be shorter probably sth like:
```ts
return TEST_CHAINS.some((chain) => chain.chainId == chainId)
```

#usesome

packages/core/src/helpers/chain.ts
```diff
@@ -1,95 +1,26 @@
...
+    .map((network) => network.chainId)
+    .includes(chainId)
 
-export function isLocalChain(chainId: ChainId) {
-  return LOCAL_CHAINS.includes(chainId)
-}
+export const isLocalChain = (chainId: ChainId) =>
+  SUPPORT_NETWORKS.filter((network) => network.isLocalChain)
+    .map((network) => network.chainId)
+    .includes(chainId)
```
tt-marek, line 26: #usesome

packages/core/src/constants/chainId.ts
```diff
@@ -1,3 +1,37 @@
...
+import { xDai } from '../model/chain/xdai'
+import { Polygon, Mumbai } from '../model/chain/polygon'
+import { Theta, ThetaTestnet } from '../model/chain/theta'
+import { Moonriver } from '../model/chain/moonriver'
+import { Harmony } from '../model/chain/harmony'
+import { Palm } from '../model/chain/palm'
+import { Fantom } from '../model/chain/fantom'
+import { Avalanche } from '../model/chain/avalanche'
+
+export const SUPPORT_NETWORKS = [
```
tt-marek, line 13: Let's rename to `DEFAULT_SUPPORTED_CHAINS`

packages/core/src/model/config/default.ts
```diff
@@ -1,29 +1,10 @@
...
-    ChainId.Mumbai,
-    ChainId.Theta,
-    ChainId.ThetaTestnet,
-    ChainId.Harmony,
-    ChainId.Moonriver,
-    ChainId.Palm,
-    ChainId.Fantom,
-    ChainId.Avalanche,
-  ],
+  supportedChains: SUPPORT_NETWORKS.map((network) => network.chainId),
```
tt-marek, line 7: Let's use here `Chain[]` instead of `number[]`.
It will allow to add custom chains.

tt-marek:
❤️ ❤️ ❤️ 
@gasolin This is very promising PR. It require a bit of work, but it will solve a lot of problems around configuration.
#393 #346 #47 
Added comments. Looking forward to merge it.

packages/core/src/constants/chainId.ts
gasolin: Make sense, but it will break the backward compatibility for DAppProvider? Tend to scope this PR to refractory only (and no test change is required to scope the impact).

Does  the suggest `getNetworkData(chainId)` is used to return the full Chain config?

packages/core/src/model/config/default.ts
gasolin: I choose not to change the default in this PR to scope the impact. 
prefer to do the breaking change in future improvement.

packages/core/src/constants/chainId.ts
tt-marek: Good point, it will break backward compatibility we should probably have a new name for it.
It is a good idea to keep scope of this PR to refactoring only, let's do that  👍 

Yes I would suggest to use getNetworkData(chainId) is used to return the full Chain config.

packages/core/src/model/config/default.ts
tt-marek: 👍

tt-marek: Let's do it in another PR, and have new name, e.g. `supportedNetworks`

gasolin:
@tt-marek PR rebased. 

- rename `MULTICALL_ADDRESS` to `multicallAddress` in per chain prop definition
- generate default multicall addresses from DEFAULT_SUPPORTED_CHAINS
- to scope the impact the refactor work neither expose new functions nor change tests. Can add `getNetworkData(chainId)` or `getMulticallAddressFor(chainId)` in future work with test, it will be much easier to do so after landing this PR

packages/core/src/constants/chainId.ts
```diff
@@ -1,3 +1,40 @@
+import { Localhost, Hardhat } from '../model/chain/local'
+import { Avalanche } from '../model/chain/avalanche'
+import { BSC, BSCTestnet } from '../model/chain/bsc'
+import { Fantom } from '../model/chain/fantom'
+import { Harmony } from '../model/chain/harmony'
+import { Mainnet, Ropsten, Rinkeby, Goerli, Kovan } from '../model/chain/ethereum'
+import { Moonriver, MoonbaseAlpha } from '../model/chain/moonriver'
+import { Palm } from '../model/chain/palm'
+import { Polygon, Mumbai } from '../model/chain/polygon'
+import { Songbird } from '../model/chain/songbird'
```
gasolin, line 10: organize imports in rough alphabet order (local chain always at the top)
to sparse the conflict when adding a new chain

packages/core/src/constants/chainId.ts
```diff
@@ -1,3 +1,40 @@
...
+export const DEFAULT_SUPPORTED_CHAINS = [
+  Localhost,
+  Hardhat,
+  Avalanche,
+  Mainnet,
+  Ropsten,
+  Rinkeby,
+  Goerli,
+  Kovan,
+  BSC,
```
gasolin, line 23: rough alphabet order, put network from the same chain together

packages/core/src/constants/chainId.ts
```diff
@@ -1,3 +1,40 @@
...
+  Palm,
+  Polygon,
+  Mumbai,
+  Songbird,
+  Theta,
+  ThetaTestnet,
+  xDai,
+]
+
 export enum ChainId {
```
gasolin, line 38: seems not possible to dynamically generate enum, which may need some attention in future work.

packages/core/src/constants/chainId.ts
tt-marek: 👍

packages/core/src/constants/chainId.ts
tt-marek: 👍

packages/core/src/constants/chainId.ts
tt-marek: Possible solution (in the following PR):
```ts
export type ChainId = KnownChainId | number
export enum KnownChainId {
  ...
}
```

packages/core/src/helpers/chain.ts
```diff
@@ -1,103 +1,22 @@
...
-      return `https://ftmscan.com/address/${address}`
-    case ChainId.Avalanche:
-      return `https://snowtrace.io/address/${address}`
-    case ChainId.Songbird:
-      return `https://songbird-explorer.flare.network/address/${address}`
-    case ChainId.MoonbaseAlpha:
-      return `https://moonbase.moonscan.io/address/${address}`
-  }
+  const networks = DEFAULT_SUPPORTED_CHAINS.filter((network) => network.chainId === chainId)
+  return networks[0]?.getExplorerAddressLink(address) || ''
```
tt-marek, line 5: Shorter version:
```ts
export function getExplorerAddressLink(address: string, chainId: ChainId) {
  return getChainById(chainId).?getExplorerAddressLink(address) || ''
 }
...
export function getChainById(chainId: ChainId) {
  DEFAULT_SUPPORTED_CHAINS.find((network) => network.chainId === chainId)
}
```

tt-marek: #getChainById

packages/core/src/helpers/chain.ts
```diff
@@ -1,103 +1,22 @@
...
-      return `https://ftmscan.com/tx/${transactionHash}`
-    case ChainId.Avalanche:
-      return `https://snowtrace.io/tx/${transactionHash}`
-    case ChainId.Songbird:
-      return `https://songbird-explorer.flare.network/tx/${transactionHash}`
-    case ChainId.MoonbaseAlpha:
-      return `https://moonbase.moonscan.io/tx/${transactionHash}`
-  }
+  const networks = DEFAULT_SUPPORTED_CHAINS.filter((network) => network.chainId === chainId)
+  return networks[0]?.getExplorerTransactionLink(transactionHash) || ''
```
tt-marek, line 10: #getChainById

packages/core/src/helpers/chain.ts
```diff
@@ -1,103 +1,22 @@
...
 }
 
 export function getChainName(chainId: ChainId) {
-  return CHAIN_NAMES[chainId]
+  const network = DEFAULT_SUPPORTED_CHAINS.find((network) => network.chainId === chainId)
+  return network?.chainName || ''
```
tt-marek, line 15: #getChainById

gasolin:
updated

tt-marek (approved):

tt-marek:
@gasolin Great work! ❤️ ❤️ ❤️ 
If you would like to continue, take a look at #408

PR #401, opened 2021-11-19
🛷 Add useTokenList

HubertNastaly:


packages/core/package.json
```diff
@@ -36,9 +36,11 @@
   },
   "dependencies": {
     "@types/react": "17.0.1",
+    "@uniswap/token-lists": "^1.0.0-beta.27",
     "@web3-react/core": "6.1.1",
     "@web3-react/injected-connector": "6.0.7",
     "@web3-react/network-connector": "6.1.5",
+    "axios": "^0.24.0",
```
HubertNastaly, line 43: Don't know if I could add axios or just use fetch

packages/core/src/hooks/useTokenList.ts
```diff
@@ -0,0 +1,33 @@
...
+import { TokenInfo } from '@uniswap/token-lists'
+import { ChainId } from '../constants'
+
+interface TokenList {
+  name: string
+  logoURI: string
+  tokens: TokenInfo[]
+}
+
+export function useTokenList(tokenListURI: string, chainId: ChainId = ChainId.Mainnet) {
```
tt-marek, line 12: I propose default chainId to be taken from currently connected network, that would be sth like this:
```ts
export function useTokenList(tokenListURI: string, overrideChainId: ChainId = null) {
  const {chainId} = useEthers()
  if (!overrideChainId) {
     overrideChainId = chainId
  } 

....
          tokens: (tokens as TokenInfo[]).filter((token) => token.chainId === overrideChainId),
...
  }, [tokenListURI, overrideChainId])
...
```

tt-marek: If chainId undefined/null - return empty list

packages/core/package.json
tt-marek: I would propose to use fetch indeed - less dependencies is better

tt-marek:
I guess documentation is missing :)

packages/core/src/hooks/useTokenList.ts
```diff
@@ -0,0 +1,33 @@
...
+          name,
+          logoURI,
+          tokens: (tokens as TokenInfo[]).filter((token) => token.chainId === chainId),
+        })
+      })
+      .catch((err) => {
+        console.log(err)
+        setTokenList(undefined)
+      })
+  }, [tokenListURI, chainId])
```
nezouse, line 30: Could just use plain fetch with checking if response is ok 
```js
.fetch(url)
    .then(async response => {
      if (response.ok) {
        return await response.json()
      } else {
        const errorMessage = await response.text()
        return Promise.reject(new Error(errorMessage))
      }
    }).catch(err)...
```

packages/example/src/components/TokensList/TokensList.tsx
```diff
@@ -1,17 +1,15 @@
-import { formatUnits } from '@ethersproject/units'
-import uniswapToken from '@uniswap/default-token-list'
```
tt-marek, line 2: Can we remove it now from package.json?

HubertNastaly: ✅

packages/core/src/hooks/useTokenList.ts
HubertNastaly: ✅

packages/core/src/hooks/useTokenList.ts
HubertNastaly: ✅

packages/core/package.json
HubertNastaly: ✅

packages/example/src/utils/adjustImageSource.ts
```diff
@@ -0,0 +1,3 @@
+export function adjustImageSource(src: string) {
```
tt-marek, line 1: `adjustIPFSPath`

docs/source/core.rst
```diff
@@ -442,6 +442,61 @@ Returns allowance (tokens left to use by spender) for given tokenOwner - spender
...
+
+If an error occurs ``undefined`` is returned.
+
+**Example**
+
+.. code-block:: javascript
+
+  const { name, logoURI, tokens } = useTokenList(UNISWAP_DEFAULT_TOKEN_LIST_URI) || {}
+
+  const httpSource = logoURI && logoURI.startsWith('ipfs') ? logoURI.replace('ipfs://', 'https://ipfs.io/ipfs/') : logoURI
```
HubertNastaly, line 470: Not sure if this should be there, but without it the example doesn't work and might be misleading

packages/example/src/utils/adjustImageSource.ts
HubertNastaly: ✅

packages/example/src/components/TokenList/TokenList.tsx
```diff
@@ -0,0 +1,142 @@
...
+  //             {token.logoURI && <TokenIcon src={token.logoURI} alt={`${token.symbol} logo`} />}
+  //           </TokenIconContainer>
+  //           <TokenName>{token.name}</TokenName>
+  //           <TokenTicker>{token.symbol}</TokenTicker>
+  //           {balances?.[idx] && <TokenBalance>{formatUnits(balances[idx]![0], token.decimals)}</TokenBalance>}
+  //         </TokenItem>
+  //       ))}
+  //   </List>
+  // )
+  const httpSource = logoURI && logoURI.startsWith('ipfs') ? logoURI.replace('ipfs://', 'https://ipfs.io/ipfs/') : logoURI
```
tt-marek: Comment 🆘

docs/source/core.rst
tt-marek: LGTM

tt-marek (approved):

tt-marek (approved):

Issue #408, opened 2021-11-20
Allow to add custom networks with any ChainId


tt-marek:
Continuation of #47
Related to #132

Currently it is impossible to use a new network without adding PR to useDApp.
We would like to change it by adding list of all possible networks to config, that can be modified.
To do so I propose to:

### Add new variable to `Config` - `networks`, i.e.

```ts
export const DEFAULT_CONFIG: FullConfig = {
  ...
  supportedChains: DEFAULT_SUPPORTED_CHAINS.map((network) => network.chainId),
  networks: DEFAULT_SUPPORTED_CHAINS
```  
so that

- `supportedChains` are list of chains ids that useDApp will connect without throwing error
- `networks` are Chain[] storing list of all known Chain and their configurations

User can dynamically edit config by using:

```ts
const config = useConfig()
const updateConfig = useUpdateConfig()
updateConfig({ networks: [...config, {...}])
```

Could be further develop into something with nicer syntax, i.e.
```ts
const {config, addNetwork, removeNetwork, ...} = useConfigReducer()
addNetwork({...})
```

### Replace all usages of constants to use useConfig, i.e.

`DEFAULT_SUPPORTED_CHAINS` -> `const {networks} = useConfig()`

Replace all usage in functions like:
```ts
getChainById(chainId: ChainId)
getExplorerAddressLink(address: string, chainId: ChainId)
getExplorerTransactionLink(transactionHash: string, chainId: ChainId)
getChainName(chainId: ChainId)
isTestChain(chainId: ChainId)
isLocalChain(chainId: ChainId)
etc...
```

### Change `ChainId` type, i.e.
```
export type ChainId = KnownChainId | number
export enum KnownChainId {
  ...
}
```

or simply to `number`

### Add documentation

Last but not least we would like to have a guide on how to add support for a new network:

- to your project
- issue a PR

### Notes

This will be breaking change and will require to advance version accordingly.

tt-marek:
Looking for comments @Szymx95 @nezouse @sz-piotr @gasolin 

@gasolin Would you like to implement?

tarrencev:
Hey! Thanks for the great work on useDapp. Just wanted to chime in and suggest using CAIP identifiers: https://github.com/ChainAgnostic/CAIPs. Thats the standard with the most traction at this point. Used by wallet connect, metamask, ect.

gasolin:
> ### Add new variable to `Config` - `networks`, i.e.
> ```ts
[...]

Issue #297, opened 2021-08-10
Error when useContractFunction on Binance Smart Chain Testnet ?


ppirch:
I implement my DApp with useDApp. Now I try to set config with this code.
```
const config: Config = {
  readOnlyChainId: 97,
  readOnlyUrls: {
    97: "https://data-seed-prebsc-1-s1.binance.org:8545/",
  },
  multicallAddresses: {
    97: "0xae11C5B5f29A6a25e955F0CB8ddCc416f522AF5C",
  },
  supportedChains: [97],
}
     
```

It works for useEthers, useEtherBalance, useTokenBalance, useContractCall, useTokenAllowance. Their state work normally , But when I try to useContractFunction then after create transaction. It says `Uncaught TypeError: Unsupported chain | The above error occurred in the <TransactionProvider> component:` 

How can I fix it? or contribute for add new network.

ppirch:
I try to add bsc testnet network in 
  @usedapp/core/src//helpers/chain.ts
  @usedapp/core/src/constants/chainId.ts
  @usedapp/core/src/model/config/default.ts
When I run useContractFunction again it's work but is this the right way to do that?

kunalalite:
It would be nice if someone could confirm that if the above approach is valid for operating with a new change using useDapp.

tt-marek:
BSC should work out of the box with #353

PR #411, opened 2021-11-21
🔗 replace supportedChains with networks in config

gasolin:
related to #408

Breaking changes:
- deprecated `supportedChains` props in Config
- deprecated chain helper functions
- Change ChainId type to number
- move `Chain`, `Config` types to packages/core/src/constants/types for outside access

New:
- add a new `networks: Chain[]` prop in Config to accept any Configs define in packages/core/src/model/chain (or outside Config when ready)
- expose networks with `const {networks} = useConfig()`

Misc:
- doc update
- put chain-specific tests to separate files


tt-marek:
@gasolin is on fire 🔥 🔥 🔥 
I will find time today or tomorrow to review it :)
Don't forget about emoji in description :D

gasolin:
@tt-marek any suggestions?

packages/core/src/helpers/chain.ts
```diff
@@ -3,16 +3,27 @@ import { DEFAULT_SUPPORTED_CHAINS, ChainId } from '../constants'
 export const getChainById = (chainId: ChainId) =>
   DEFAULT_SUPPORTED_CHAINS.find((network) => network.chainId === chainId)
 
-export const getExplorerAddressLink = (address: string, chainId: ChainId): string | '' =>
-  getChainById(chainId)?.getExplorerAddressLink(address) || ''
+export const getExplorerAddressLink = (address: string, chainId: ChainId): string | '' => {
+  console.warn('getExplorerAddressLink is deprecated, can call with Chain directly')
```
marekkirejczyk, line 7: Extract to separate method: 
`deprecationWarning('getExplorerAddressLink is deprecated, can call with Chain directly')`

packages/core/src/constants/type/Config.ts
```diff
@@ -12,7 +12,7 @@ export type FullConfig = {
   readOnlyChainId?: ChainId
   readOnlyUrls?: NodeUrls
   multicallAddresses?: MulticallAddresses
-  supportedChains: number[]
+  networks: Chain[]
```
marekkirejczyk, line 15: For backward compatibility I would do the following thing:
1) keep both `supportedChains` and introduce `networks`
2) Add deprecation warning to usage of `supportedChains`
3) Add custom logic to useConfig, that will populate networks if empty based `supportedChains`

packages/core/src/constants/chainId.ts
```diff
@@ -37,26 +48,4 @@ export const DEFAULT_SUPPORTED_CHAINS = [
...
-  Harmony = 1666600000,
-  Palm = 11297108109,
-  Localhost = 1337,
-  Hardhat = 31337,
-  Fantom = 250,
-  Avalanche = 43114,
-  Songbird = 19,
-  MoonbaseAlpha = 1287,
-}
+export type ChainId = number
```
marekkirejczyk, line 51: What do you think about adding type according to @sz-piotr proposal?

packages/core/test/model/chain/avalanche.test.ts
```diff
@@ -0,0 +1,29 @@
...
+  })
+
+  it('getExplorerAddressLink', () => {
+    expect(Avalanche.getExplorerAddressLink(TEST_ADDRESS)).to.eq(`https://snowtrace.io/address/${TEST_ADDRESS}`)
+  })
+
+  it('getExplorerTransactionLink', () => {
+    expect(Avalanche.getExplorerTransactionLink(TEST_TX)).to.eq(`https://snowtrace.io/tx/${TEST_TX}`)
+  })
+})
```
marekkirejczyk, line 29: Niceee work ❤️

marekkirejczyk:
@gasolin Great work. Added a few comments. This is a bit tricky refactoring, I think we should not break compatibility completely in one step. Let me know what you think about comments.

packages/core/src/model/chain/index.ts
```diff
@@ -0,0 +1,13 @@
+export * from './/local'
```
marekkirejczyk, line 1: why `//`?

packages/core/src/providers/DAppProvider.tsx
```diff
@@ -28,13 +27,10 @@ interface WithConfigProps {
...
-const MULTICALL_ADDRESSES: { [index: number]: string } = {}
-DEFAULT_SUPPORTED_CHAINS.filter((network) => network.multicallAddress).map(
-  (network) => (MULTICALL_ADDRESSES[network.chainId] = network.multicallAddress)
-)
-
 function DAppProviderWithConfig({ children }: WithConfigProps) {
-  const { multicallAddresses } = useConfig()
+  const { multicallAddresses, networks } = useConfig()
+  const MULTICALL_ADDRESSES: { [index: number]: string } = {}
+  networks.map((network) => (MULTICALL_ADDRESSES[network.chainId] = network.multicallAddress))
```
marekkirejczyk, line 33: Extract metod and cache using useMemo, sth like this

```ts
function getMulticallAddresses(networks) {
  const result: { [index: number]: string } = {}
  networks.each((network) => (result[network.chainId] = network.multicallAddress))
  return result;
}

...
const defaultAddresses = useMemo(() => getMulticallAddresses(networks));

const multicallAddressesMerged = { ...defaultAddresses, ...multicallAddresses }
```

packages/core/src/helpers/chain.ts
```diff
@@ -3,16 +3,27 @@ import { DEFAULT_SUPPORTED_CHAINS, ChainId } from '../constants'
...
+  return getChainById(chainId)?.getExplorerAddressLink(address) || ''
+}
 
-export const getExplorerTransactionLink = (transactionHash: string, chainId: ChainId): string | '' =>
-  getChainById(chainId)?.getExplorerTransactionLink(transactionHash) || ''
+export const getExplorerTransactionLink = (transactionHash: string, chainId: ChainId): string | '' => {
```
marekkirejczyk, line 11: As we deprecating all those methods, I think we should introduce new hook, like useNetwork() that will return current network

packages/core/test/model/chain/avalanche.test.ts
marekkirejczyk: The tests for each network in individual file look great!

packages/core/src/model/chain/index.ts
gasolin: typo 😅

packages/core/src/helpers/chain.ts
gasolin: since `getChainById` still works through DEFAULT_SUPPORTED_CHAINS, can add so in the next major version

marekkirejczyk: 👍

packages/core/src/constants/type/Config.ts
```diff
@@ -12,7 +12,8 @@ export type FullConfig = {
   readOnlyChainId?: ChainId
   readOnlyUrls?: NodeUrls
   multicallAddresses?: MulticallAddresses
-  supportedChains: number[]
+  supportedChains?: number[]
+  networks?: Chain[]
```
marekkirejczyk, line 16: Looks good, I would add in [useConfig](https://github.com/EthWorks/useDApp/blob/cdb02b9a260b99d85622401a83037bcea07c8a02/packages/core/src/providers/config/context.ts#L10) sth like:
```ts
if (config.supportedChains) {
  const networks = getNetworksByIds(config.supportedChains)
  return {...config, networks}
}   
return config
```

So that we still use supportedNetworks if someone is using it while fallback to networks if no supportedChains are present.

packages/core/src/model/config/default.ts
```diff
@@ -1,10 +1,9 @@
-import { DEFAULT_SUPPORTED_CHAINS } from '../../constants/chainId'
-
-import { FullConfig } from './Config'
+import { DEFAULT_SUPPORTED_CHAINS, FullConfig } from '../../constants'
 
 export const DEFAULT_CONFIG: FullConfig = {
   pollingInterval: 15000,
   supportedChains: DEFAULT_SUPPORTED_CHAINS.map((network) => network.chainId),
```
marekkirejczyk, line 5: we can initialise `supportedChains` to null/undefined now

gasolin:
@tt-marek comment addressed, please review again

vm06007:
looking forward, will this support useDapp to be able switching between multiple networks and provide your own network such as smartBCH?

gasolin:
> looking forward, will this support useDapp to be able switching between multiple networks and provide your own network such as smartBCH?

not yet because of backward compatibility, but will make it way easier to provide your own network in the next major version

for switching between multiple networks, could take a look https://github.com/EthWorks/useDApp/issues/416

vm06007:
no problem from reading on multiple networks and having a switch, the problem is when sending transaction (useContractFunction) gives an error same as here: https://github.com/EthWorks/useDApp/issues/297 (trying with network 10000, smartBCH)

would it be better to open PR with new chain or wait for the update so can do useContractFunction properly?

gasolin:
@marekkirejczyk any feedback for this PR?

docs/source/getting-started.rst
```diff
@@ -29,13 +29,13 @@ Below is a simple example:
 
 .. code-block:: javascript
 
-  import { ChainId, DAppProvider, useEtherBalance, useEthers, Config } from '@usedapp/core'
+  import { Mainnet, DAppProvider, useEtherBalance, useEthers, Config } from '@usedapp/core'
   import { formatEther } from '@ethersproject/units'
 
   const config: Config = {
-    readOnlyChainId: ChainId.Mainnet,
+    readOnlyChainId: Mainnet.chainID,
```
tt-marek, line 36: Shall we use now?
`networks: Mainnet`

docs/source/guide.rst
```diff
@@ -11,13 +11,14 @@ To connect to the network in read-only mode, provide ``readOnlyChainId`` and ``r
 See example configuration below:
 
 .. code-block:: javascript
- 
+  import { Mainnet } from '@usedapp/core'
+
   const config = {
-    readOnlyChainId: ChainId.Mainnet,
+    readOnlyChainId: Mainnet.chainID,
```
tt-marek, line 17: `networks: Mainnet`?

tt-marek (approved):

vm06007:
cool its merged! so now is it possible to specify own custom chain? 
if so, does this require to use `networks ` property in config? 
would love to see an example!

tt-marek:
We still need:

1. test the newest version on master (lot's of changes, sanity check)
2. add changset and publish npm
3. a quick tutorial adding a network to a project
4. a quick tutorial adding a network to useDApp

@gasolin 
Thank you for the great work! ❤️ Sorry for being slow to review. Promise I'll do better 🙏 
Would you like to take on any of 1,3,4?

tt-marek:
@gasolin I published npm, however ChainId is still enum, do you want to work on it?

gasolin:
> @gasolin Thank you for the great work! ❤️ Sorry for being slow to review. Promise I'll do better 🙏 Would you like to take on any of 1,3,4?

will check if I can do some doc improvement

PR #415, opened 2021-11-22
⚒Add typed contract hooks

Szymx95:
Screenshot 2021-12-17 at 11 52 54
Screenshot 2021-12-17 at 11 54 25
TypedContractCall will type check method and args


packages/core/src/model/types.ts
```diff
@@ -1 +1,15 @@
+/* eslint-disable */
+import { Contract, ContractTransaction } from "ethers"
+
 export type Falsy = false | 0 | '' | null | undefined
+
+export type TypedContract = Contract & {
+    functions: Record<string, (...args: any[]) => any>
+}
+
+export type ContractFunctionNames<T extends TypedContract> = keyof { [P in keyof T['functions'] as ReturnType<T['functions'][P]> extends Promise<ContractTransaction> ? P : never] : void } & string
```
Szymx95, line 10: eslint for some reason finds a syntax error even though there is none. Probably something with version mismatch. Will have to check in future PR

.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules
 dist
 .DS_Store
 .idea
+packages/example/contracts
```
tt-marek, line 5: Noise?

packages/core/.prettierignore
```diff
@@ -0,0 +1 @@
+src/model/types.ts
```
marekkirejczyk, line 1: Missing newline?

.gitignore
Szymx95: Those are files generated with typechain 

> "typechain:generate": "yarn typechain --target=ethers-v5 ./src/abi/**/*.json --out-dir=./contracts"

Propably better not to include them in repo

tt-marek, line 5: Proposed directory: `packages/example/gen/types`

packages/example/package.json
```diff
@@ -3,8 +3,9 @@
   "version": "0.2.10",
   "private": true,
   "scripts": {
+    "typechain:generate": "yarn typechain --target=ethers-v5 ./src/abi/**/*.json --out-dir=./contracts",
```
tt-marek, line 6: -> `gen/types`

docs/source/core.rst
```diff
@@ -107,6 +107,49 @@ It is recommended to use `useContractCall`_ where applicable instead of this met
 
 - ``calls: ChainCall[]`` - list of calls, also see `ChainCall`_. Calls need to be in the same order across component renders.
 
+.. _useTypedContractCall-label:
+
+useTypedContractCall
```
tt-marek, line 112: Can we merge `useTypedContractCall` and `useContractCall` into one function that accepts to different Call types as arguments?

Issue #432, opened 2021-12-08
Module not found: Can't resolve 'lodash/merge'


jaaneh:
Hello!

Very eager to test out this library, but I'm met with an error when compiling. I can see #321 #323 with similar issues.

Can this be since I'm using Next.js? The example in your repo works fine with normal React.

```js
./node_modules/@usedapp/core/dist/esm/src/providers/config/reducer.js:1:0
Module not found: Can't resolve 'lodash/merge'

Import trace for requested module:
./node_modules/@usedapp/core/dist/esm/src/providers/config/provider.js
./node_modules/@usedapp/core/dist/esm/src/providers/DAppProvider.js
./node_modules/@usedapp/core/dist/esm/src/providers/index.js
./node_modules/@usedapp/core/dist/esm/src/index.js
./src/pages/_app.tsx

https://nextjs.org/docs/messages/module-not-found
```

Current _app.tsx setup:
```js
import * as React from 'react'
import { AppProps } from 'next/app'
import { ThemeProvider } from '@mui/material/styles'
import CssBaseline from '@mui/material/CssBaseline'
import { CacheProvider, EmotionCache } from '@emotion/react'
import { DAppProvider, Mainnet } from '@usedapp/core'

import theme from '../styles/theme'
import createEmotionCache from '../utils/createEmotionCache'

const clientSideEmotionCache = createEmotionCache()

const dappConfig = {
  readOnlyChainId: Mainnet.chainId,
  readOnlyUrls: {
    [Mainnet.chainId]: process.env.NEXT_PUBLIC_INFURA_KEY
  }
}

interface MyAppProps extends AppProps {
  emotionCache?: EmotionCache
}

const MyApp = (props: MyAppProps) => {
  const { Component, emotionCache = clientSideEmotionCache, pageProps } = props

  return (
    <DAppProvider config={dappConfig}>
      <CacheProvider value={emotionCache}>
        <ThemeProvider theme={theme}>
          <CssBaseline />
          <Component {...pageProps} />
        </ThemeProvider>
      </CacheProvider>
    </DAppProvider>
  )
}

export default MyApp
```

jaaneh:
Seems like we're still loading `lodash/merge` in master [config/reducer.ts](https://github.com/EthWorks/useDApp/blob/master/packages/core/src/providers/config/reducer.ts) while `lodash.merge` was merged in #323 🤔

PR #436, opened 2021-12-13
Fix can't resolve 'lodash/merge #432

gasolin:
fix regression of #411


gasolin:
@marekkirejczyk please review the fix

marekkirejczyk (approved):

PR #465, opened 2022-01-05
🧿 update node matrix in CI

marcin-trust:
Align with current LTS support

12, 14, 16 are LTS versions
https://nodejs.org/en/about/releases/

### DIFF
--- a/.changeset/smart-dragons-admire.md
+++ b/.changeset/smart-dragons-admire.md
@@ -1,0 +1,9 @@
+---
+'@usedapp/coingecko': minor
+'@usedapp/core': minor
+'@usedapp/extension': minor
+'@usedapp/testing': minor
+'@usedapp/example': minor
+---
+
+Support Node LTS v14 and v16, abandon v10 and v12
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -8,7 +8,7 @@ jobs:
   test:
     strategy:
       matrix:
-        node: ['10.x', '12.x']
+        node: ['14.x', '16.x']
         os: [ubuntu-latest]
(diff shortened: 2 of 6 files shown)


jodator:
@marcin-trust It looks like the `test 16` was canceled. I'm re-running the jobs to see if this was a GH hickup.

marcin-trust:
@jodator at the same time you need to reconfigure branch policy and remove 10.x and 12.x required checks

jodator:
> @jodator at the same time you need to reconfigure branch policy and remove 10.x and 12.x required checks

Yep, we will need to do this after the merge. For now, it looks like the job on node `16` is running for too long. Could check why is that?

marcin-trust:
node v16 test run hangs locally. why-is-node-running prints forever in async hook loop or sth.
I added `--exit` to mocha and I'll create an issue

marcin-trust:
#476 related

marcin-trust:
> I added --exit to mocha and I'll create an issue

It didn't help

jodator:
> It didn't help

@marcin-trust Could you check to add `--exit` flag to other packages? I've tested this locally, and I had to add this flag to all packages.

marcin-trust:
@jodator - yes, it helped, let's see in this PR

marcin-trust:
I forgot about testing project. Now it should work in CI too

marcin-trust:
It got passed. Now you @jodator need to update the required check rules.

jodator:
@marcin-trust I think that the last step is also to include a change set as we introduce a breaking change. Since we're still on `0.x` I'd go with MINOR. Other than that, we're good to go.

jodator (approved):

PR #472, opened 2022-01-07
Add resetState function for transactions

DoubleOTheven:
* Return `resetState` in `usePromiseTransaction` so that it is easy to reset the view after a transaction.
* Return `resetState` in `useContractFunction`
* Return `resetState` in `useSendTransaction`

Currently there is no way to reset the UI after a user has submitted a transaction (that results in `Success`, `Fail`, or `Exception`). This PR makes it possible to call `resetState()` to reset the state to `None`. Let's say that as a developer we check for the `Success` status so we know we need to show a success React component. In this component we also have a "Mint More NFTs" button that we want to reset the status to `None` when clicked so that we can show the `MintNFTForm` React component. We can add `onClick={resetState}` to our button on the success React component and all of our problems are solved!

### DIFF
--- a/.changeset/moody-candles-thank.md
+++ b/.changeset/moody-candles-thank.md
@@ -1,0 +1,5 @@
+---
+'@usedapp/core': minor
+---
+
+Add resetState function to useContractFunction and useSendTransaction so the UI can be reset easily after a transaction attempt.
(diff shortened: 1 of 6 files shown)


DoubleOTheven:
@tt-marek Mind taking a look at this one? This change would solve many real-world problems. I'm open to feedback

DoubleOTheven:
This is a very useful feature for real-world apps. Mind taking a look? @marekkirejczyk

tt-marek (approved):

tt-marek:
Great job! Thanks for fixing all the typos.

PR #474, opened 2022-01-10
🧐 Replace polygon explorer to polyscan

gasolin:
polygon's default explorer is changed to https://polygonscan.com/


jodator (approved):
Thanks!

And double thanks for adding emoji to the PR's title 🎉

PR #475, opened 2022-01-10
✨ add gnosis chain as xdai alias

gasolin:
refer https://www.xdaichain.com/

xdai chain has rename to gnosis chain

Due to lots of user use the legacy xdai name, tend to keep it and just add the alias


tt-marek:
This is compatibility breaking change, can we have both names?

gasolin:
added back

tt-marek:
Needs. rebase + conflict solving

gasolin:
rebased

tt-marek (approved):

PR #479, opened 2022-01-12
🥶 Bump web3-react to 8.0.0-beta, add Web3Modal to example

vanruch:
## What has changed
**The good:**
- Updated web3React to beta version
- Prepare for simulations multichain support
- Support Web3Modal
- Better error handling
- Got rid of deprecation console warnings
- (Almost) no breaking changes

**The bad**
- web3-react 8 is an ESM module which causes problems with ts-node and mocha. I made it work somehow by making useDapp esm module and changing tsconfig
- web3-react is in beta and is still in active development, so if we go on with this version, we'll have to introduce useDapp beta version and develop it in parallel with production

**The ugly**
- E2E tests aren't working and I have no idea why

Edit: fixed E2E tests


Marik-D (approved):
Older versions of @web-react seem to be installed as well. Maybe theres a dependency that wasn't upgraded?

PR #484, opened 2022-01-14
🦎 Support metis mainnet

titandac:
add metis mainnet and rename Stardust to MetisStardust to make it more clear it's a Metis testnet.


packages/core/src/constants/chainId.ts
```diff
@@ -80,7 +82,8 @@ export enum ChainId {
   Avalanche = 43114,
   Songbird = 19,
   MoonbaseAlpha = 1287,
-  Stardust = 588,
+  Metis = 1088,
+  MetisStardust = 588,
```
tt-marek, line 86: That is compatibility breaking change, better stay with Stardust...

titandac: that's a good point. new commit incoming.

titandac: I changed it back to 'Stardust'. Additionally I think it makes more sense to refer to the mainnet chain by it's official name, 'Andromeda'. My understanding is that in the future there may be multiple Metis mainnet chains, therefore calling one of them 'Metis' may become misleading.

tt-marek:
Need lint fix

titandac:
thanks for waiting on this, wasn't able to work on this until today.

fixed the lint issue and squashed into a single commit.

mj426382 (approved):

PR #489, opened 2022-01-18
🥶 Bump web3-react to 8.0.0-beta, add Web3Modal to example (#479)

vanruch:


tt-marek:
Will continue in #491 with alternative solution.

PR #495, opened 2022-01-21
🤙 Return error from useContractCall

HubertNastaly:
In current useContractCall implementation we get `undefined` in both cases: when data is not fetched yet or when an error occurs. This PR will provide the functionality to distinguish these cases.

### DIFF
--- a/.changeset/slimy-baboons-design.md
+++ b/.changeset/slimy-baboons-design.md
@@ -1,0 +1,6 @@
+---
+"@usedapp/core": minor
+"@usedapp/example": minor
+---
+
+🤙 Return error from useContractCall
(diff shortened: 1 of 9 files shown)


nezouse:
Did you test if messages are anyhow meaningful? Probably on ganache they will be different, but would be good to check that.

marekkirejczyk (changes requested):
As discussed - introduce new function `useCall` to avoid compatibility breaking changes

nezouse:
If introduced as new call in same PR please change the changeset as i marked it as minor, because it was breaking change.

Tono1852 (changes requested):
@

PR #501, opened 2022-01-25
📱 Support and add examples of Web3Modal and Web3React provider usages

vanruch:


packages/core/src/providers/network/provider.tsx
```diff
@@ -1,14 +1,30 @@
...
+  try {
+    return await provider.getSigner().getAddress()
+  } catch (e) {
+    if (e.code === 'UNSUPPORTED_OPERATION') {
+      // readonly provider
+      return undefined
+    }
+    throw e
+  }
+}
```
tt-marek, line 23: 👍

tt-marek: Ha ha, I have a next question for you about error handling: Should we throw error, or add error to state?

tt-marek (approved):

packages/core/src/providers/network/provider.tsx
vanruch: It will call reportError after it's caught in `activate` function. Btw this is still not obvious if we want it - a fail fast approach might work better here and we should throw when somebody tries to activate bad provider

PR #504, opened 2022-01-28
📧 Add useChainMeta hook to get chain metadata from ChainID

vanruch:


packages/core/src/hooks/useChainMeta.ts
```diff
@@ -0,0 +1,6 @@
+import * as chains from '../model/chain'
+import { ChainId } from '../constants'
+
+export function useChainMeta(chainId: ChainId) {
```
tt-marek, line 4: Why not getChainMeta? I does not seem to be a hook.

vanruch: Technically not but this will be more in style with the rest of usedapp so I made it look like a hook.

tt-marek: Well one expect hook to somehow be connected to component refresh, so in that case, it would break the convention. However if we add useMemo to reduce amount of computations...

tt-marek (approved):

PR #517, opened 2022-02-02
🧋 Add Boba support

Quantumlyy:


mj426382 (commented):
Everything looks good and also tests work.
I verified boba multicall address, can You verify testnet one?
I am not able to check the code of this :(

PR #558, opened 2022-03-08
🧐 Improve testing infrastructure

dmaretskyi:


packages/testing/package.json
```diff
@@ -22,20 +22,20 @@
...
     "prettier": "^2.1.2",
-    "typescript": "^4.1.3"
+    "typescript": "^4.6.2"
   },
   "dependencies": {
     "@types/react": "17.0.1",
     "@types/react-dom": "17.0.1",
     "@usedapp/core": "*",
     "ethers": "5.5.1",
-    "ethereum-waffle": "3.2.2",
```
rzadp, line 32: Why? It's a `peer` of `core`, will it work when `@usedapp/testing` is installed from NPM?

PR #561, opened 2022-03-08
🦔 Alternative Switch to PNPM

rzadp:
Alternative to https://github.com/TrueFiEng/useDApp/pull/560 using `pnpm import`


packages/core/src/providers/network/provider.tsx
```diff
@@ -51,7 +52,10 @@ export function NetworkProvider({ children }: NetworkProviderProps) {
 
   const activate = useCallback(
     async (provider: JsonRpcProvider | (EventEmitter & ExternalProvider)) => {
-      const wrappedProvider = provider instanceof JsonRpcProvider ? provider : new Web3Provider(provider)
+      const wrappedProvider =
+        provider instanceof JsonRpcProvider || provider instanceof providers.Web3Provider
```
dmaretskyi, line 56: Can it an already existing import from `@ethersproject/providers`? The `ethers` one should be just a re-export

rzadp: It doesn't work when I try that.

dmaretskyi: why?

rzadp: It's not that instance apparently, it gets wrapped by `new Web3Provider(provider)` and then it crashes with weird errors without stacktracke:

```
Error: Method [object Object] not supported.
```

dmaretskyi: looks like there are duplicate dependencies installed 
ethers@5.5.0 vs ethers@5.5.3

dmaretskyi (approved):

PR #577, opened 2022-03-11
🔗 Explicitly export public symbols 

dmaretskyi:


packages/core/src/index.ts
```diff
@@ -1,5 +1,136 @@
-export * from './constants'
-export * from './providers'
-export * from './hooks'
-export * from './model'
-export * from './helpers'
+export {
```
rzadp, line 1: Add a note why we export explicitly?

rzadp (approved):

PR #581, opened 2022-03-12
docs: use "import type" in getting started example

paulrberg:


paulrberg:
Why was this PR closed?

PR #591, opened 2022-03-15
🔗 Improve WalletConnect example

alexworker23:
## Description of changes
Extend WalletConnect example by adding:
- Disconnect functionality
- Added "Sign" button for signing message testing

## Notes 
Also suggest 1 improvement to "Send Ether", "Balance", "Tokens", "Transactions" pages: 
- Remove "Connect" button from the header of those pages so that wallet connection will be handled by appropriate pages (Web3Modal, WalletConnect, Web3ReactConnector, etc.). 
It will make those 4 pages to serve just as playgrounds for blockchain interactions. 
Currently "Connect" button on those pages handles only Metamask connection.


dmaretskyi (approved):

packages/core/src/hooks/useEthers.ts
```diff
@@ -45,12 +45,16 @@ export function useEthers(): Web3Ethers {
...
-        return activate(await providerOrConnector.getProvider())
+      try {
+        if ('getProvider' in providerOrConnector) {
+          console.warn('Using web3-react connectors is deprecated and may lead to unexpected behavior.')
+          await providerOrConnector.activate()
+          return activate(await providerOrConnector.getProvider())
+        }
+        return activate(providerOrConnector)
+      } catch (error) {
+        console.error(error)
```
dmaretskyi, line 56: I don't think we should catch the errors in activate but rather allow the user code to handle them

alexworker23: True, removed error catching

dmaretskyi (approved):

PR #592, opened 2022-03-15
🍡 Add multichain test

mj426382:


packages/core/src/hooks/multiChainCalls.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Contract } from '@ethersproject/contracts'
+import { useChainCall } from '..'
+import { expect } from 'chai'
+import { renderWeb3Hook, deployMockToken, MOCK_TOKEN_INITIAL_BALANCE } from '../testing'
+import { encodeCallData } from '../helpers'
+import { ChainId } from '../constants/chainId'
+
+const secondTestChainId = 2137
+
+describe.only('multiChainCalls', () => {
```
dmaretskyi, line 11: Remove .only

packages/core/src/hooks/multiChainCalls.test.tsx
```diff
@@ -0,0 +1,52 @@
+import { MockProvider } from '@ethereum-waffle/provider'
+import { Contract } from '@ethersproject/contracts'
+import { useChainCall } from '..'
+import { expect } from 'chai'
+import { renderWeb3Hook, deployMockToken, MOCK_TOKEN_INITIAL_BALANCE } from '../testing'
+import { encodeCallData } from '../helpers'
+import { ChainId } from '../constants/chainId'
+
+const secondTestChainId = 2137
```
dmaretskyi, line 9:
```suggestion
const SECOND_TEST_CHAIN_ID = 2137
```

packages/core/src/testing/renderWeb3Hook.tsx
```diff
@@ -8,6 +8,7 @@ import { ReadonlyNetworksProvider } from '../providers/network'
 
 export interface renderWeb3HookOptions<Tprops> {
   mockProvider?: MockProvider
+  otherProvider?: MockProvider
```
dmaretskyi, line 11: I would make it more generic:

```typescript
mockProvider?: MockProvider | Record<number /* ChainId */, MockProvider>
```

so the provider passed in is either a single provider or a map of `chainId => provider` if you want multiple.

packages/core/src/hooks/multiChainCalls.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+              args,
+            },
+            chainId
+          )
+        ),
+      {
+        mockProvider,
+        otherProvider: secondMockProvider,
+      }
+    )
```
dmaretskyi, line 46:
```suggestion
    const { result, waitForCurrent } = await renderWeb3Hook(
      () => useChainCall(encodeCallData(
            {
              contract,
              method: 'balanceOf',
              args,
            },
            chainId
      )),
      {
        mockProvider,
        otherProvider: secondMockProvider,
      }
    )
```

packages/core/src/hooks/multiChainCalls.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+    await waitForCurrent((val) => val !== undefined)
+    expect(result.error).to.be.undefined
+    expect(result.current).to.eq(MOCK_TOKEN_INITIAL_BALANCE)
```
dmaretskyi, line 50: Both tests have exactly the same assertions: the balance on both chains is the same. So even if the code wasn't working the tests might pass. Can you make so that different chains return different values?

packages/core/src/hooks/multiChainCalls.test.tsx
dmaretskyi, line 11: I would move those tests into the test suites for specific hook. For example this would be in `useChainCall.test.ts`

packages/core/src/hooks/multiChainCalls.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  it('useChainCall to return correct values', async () => {
+    await testUseCustomCallSingleChainFromMulti(token, [deployer.address], ChainId.Localhost)
+    await testUseCustomCallSingleChainFromMulti(secondToken, [secondDeployer.address], secondTestChainId)
+  })
+
+  const testUseCustomCallSingleChainFromMulti = async (contract: Contract, args: string[], chainId: number) => {
+    const { result, waitForCurrent } = await renderWeb3Hook(
+      () =>
+        useChainCall(
```
dmaretskyi, line 32: useChainCall is getting deprecated and replaced by useRawCall. Can you add a test for it as well?

packages/core/src/hooks/multiChainCalls.test.tsx
```diff
@@ -0,0 +1,52 @@
+import { MockProvider } from '@ethereum-waffle/provider'
```
dmaretskyi, line 1: Would also be good if you added similar small multi-chain tests for easy hooks like useEtherBalance

dmaretskyi (approved):

dmaretskyi (changes requested):
Change requests in comments

packages/core/src/hooks/multiChainCalls.test.tsx
mj426382: Unfortunately it is linter change:(

dmaretskyi (approved):

PR #603, opened 2022-03-16
🔥 Updated useTransaction status values

mateomolinari:
current version of useTransaction uses 'Fail' as tx status, instead of 'Failed'


rzadp (approved):
Thanks.

PR #609, opened 2022-03-18
🥳 Move all examples to embeddables

mj426382:


packages/docs/src/examples/EthBalance.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+ReactDOM.render(
+    <DAppProvider config={config}>
+      <EthBalance />
+    </DAppProvider>,
+    document.getElementById('root')
+)
+
+export function EthBalance() {
```
rzadp, line 20: We name every example component `App` because every example is a standalone app.
The imports won't collide because you have default exports, so you can still import it as `EthBalance` in MDX.

rzadp (changes requested):

packages/docs/src/examples/EthBalance.tsx
mj426382: Fine, i revert my changes

rzadp (approved):

PR #614, opened 2022-03-21
⚧ Add transaction example to docs

mj426382:


packages/docs/src/examples/Transaction.tsx
```diff
@@ -0,0 +1,59 @@
...
+    }, [state])
+
+    useEffect(() => {
+        setDisabled(chainId === ChainId.Mainnet)
+    }, [chainId])
+
+    const { activateBrowserWallet, account } = useEthers()
+
+    return (
+    <div>
```
dmaretskyi, line 45: formatting

packages/docs/src/examples/Transaction.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <button onClick={() => activateBrowserWallet()}>Connect</button>
+        {account && <p>Account: {account}</p>}
+        {disabled 
+        ? (<p>Change network from mainnet</p>)
```
dmaretskyi, line 49:
```suggestion
        ? <p>Change network from mainnet</p>
```

packages/docs/src/examples/Transaction.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+    const send = () => {
+        sendTransaction({ to: address, value: parseEther(amount) })
+    }
+
+    useEffect(() => {
+        setStatus(state.status)
+    }, [state])
+
+    useEffect(() => {
```
dmaretskyi, line 38: can it be just ?

```
const disabled = chainId === ChainId.Mainnet
```

mj426382: Of course it can be, i thought that i need to make 1:1 with upper example

packages/docs/src/examples/Transaction.tsx
```diff
@@ -0,0 +1,51 @@
...
+        : (
+        <div>
+            <button onClick={() => send()}>Send ether</button>
+            <p>Status: {status}</p>
+        </div>
+        )
+        }
+    </div>
+    )
+}
```
rzadp, line 51: newline

packages/docs/src/examples/Transaction.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+    const send = () => {
+        sendTransaction({ to: address, value: parseEther(amount) })
+    }
+
+    const { activateBrowserWallet, account } = useEthers()
+
+    return (
+    <div>
+        <button onClick={() => activateBrowserWallet()}>Connect</button>
```
rzadp, line 38: Shall we make the button disappear when we have already connected?
(Applies to every other example as well)

packages/docs/docs/guides/transactions.mdx
```diff
@@ -43,6 +41,16 @@ Example below clears inputs and enables all disabled components back:
     }, [state])
 ```
 
+**Live example**
+
+App will send 0.5 ether to a test account.
+Connect a MetaMask wallet and switch to a test network.
```
rzadp, line 47:
```suggestion
Connect a MetaMask wallet **and switch to a test network**, such as Kovan or Ropsten.
```

packages/docs/docs/guides/transactions.mdx
```diff
@@ -43,6 +41,16 @@ Example below clears inputs and enables all disabled components back:
     }, [state])
 ```
 
+**Live example**
+
+App will send 0.5 ether to a test account.
```
rzadp, line 46: Why so much? It's free ether but still in my experience testnet ether is not always easy to get when faucets are dry.
Let's make it 1 Gwei.

packages/docs/src/examples/Transaction.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+ReactDOM.render(
+    <DAppProvider config={config}>
+      <TokenBalance />
+    </DAppProvider>,
+    document.getElementById('root')
+)
+
+export function TokenBalance() {
```
rzadp, line 21: copy pasted name

packages/docs/src/examples/Transaction.tsx
```diff
@@ -0,0 +1,51 @@
...
+    const { chainId } = useEthers()
+    const { sendTransaction, state } = useSendTransaction()
+    
+    const status = state.status
+    const disabled = chainId === ChainId.Mainnet
```
rzadp, line 26:
```suggestion
    // We prevent the example from running on Mainnet so that the users do not use real Ether without realizing.
    const disabled = chainId === ChainId.Mainnet
```

Always good to provide some explanation for non-obvious things.

packages/docs/src/examples/Transaction.tsx
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    const { activateBrowserWallet, account } = useEthers()
+
+    return (
+    <div>
+        <button onClick={() => activateBrowserWallet()}>Connect</button>
+        {account && <p>Account: {account}</p>}
+        {disabled
+        ? <p>Change network from mainnet</p>
```
rzadp, line 41:
```suggestion
        ? <p>Please change the network from Mainnet to proceed.</p>
```

packages/docs/src/examples/Transaction.tsx
```diff
@@ -0,0 +1,51 @@
...
+    const send = () => {
+        sendTransaction({ to: address, value: parseEther(amount) })
+    }
+
+    const { activateBrowserWallet, account } = useEthers()
+
+    return (
+    <div>
+        <button onClick={() => activateBrowserWallet()}>Connect</button>
+        {account && <p>Account: {account}</p>}
```
rzadp, line 39: I propose we also hide it when `disabled`, so that the user sees only the call to action to change the network.

packages/docs/src/examples/Transaction.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react'
+import ReactDOM from 'react-dom'
+import { parseEther } from '@ethersproject/units'
+import { Mainnet, DAppProvider, useSendTransaction, useEthers, Config } from '@usedapp/core'
+import { ChainId } from '../../../core/src/constants/chainId'
```
rzadp, line 5: Watch it, a bad import. It won't work for the users.

rzadp (changes requested):

packages/docs/src/examples/Transaction.tsx
mj426382: I'll make task for that and i can fix that, also from me is to export all configs except for getting started to the other file. What do You think?

rzadp: We can give it a try.

packages/docs/src/examples/Transaction.tsx
mj426382: What is the funniest, it worked 😂

packages/docs/src/examples/Transaction.tsx
```diff
@@ -0,0 +1,48 @@
...
+    const { chainId } = useEthers()
+    const { sendTransaction, state } = useSendTransaction()
+    
+    // We prevent the example from running on Mainnet so that the users do not use real Ether without realizing.
+    const disabled = chainId === ChainId.Mainnet
+    const status = state.status
+    const address = '0xe13610d0a3e4303c70791773C5DF8Bb16de185d1'
+
+    const send = () => {
+        sendTransaction({ to: address, value: 1 })
```
rzadp, line 29: This is one `wei`, not `Gwei` :D So either change it here or in the description of the example.

mj426382: :)

packages/docs/src/examples/Transaction.tsx
```diff
@@ -0,0 +1,54 @@
...
+        </div>
+        )
+    }
+
+    return (
+    <div>
+        {!account && <button onClick={() => activateBrowserWallet()}>Connect</button>}
+        {account 
+        ? <WalletContent />
+        : <p>Connect to wallet to integrate with example</p>
```
rzadp, line 50:
```suggestion
        : <p>Connect to wallet to interact with the example.</p>
```

rzadp (approved):

PR #616, opened 2022-03-21
Instantiate readonly providers on-demand

dmaretskyi:


packages/core/src/providers/network/readonlyNetworks/provider.tsx
```diff
@@ -16,14 +15,45 @@ export const getProvidersFromConfig = (readOnlyUrls: NodeUrls) =>
 
 export function ReadonlyNetworksProvider({ providerOverrides = {}, children }: NetworkProviderProps) {
   const { readOnlyUrls = {} } = useConfig()
-  const [providers, setProviders] = useState<Providers>(() => ({
-    ...getProvidersFromConfig(readOnlyUrls),
-    ...providerOverrides,
-  }))
+
```
dmaretskyi, line 18: prehaps store requested chainIds in a separate state

PR #638, opened 2022-03-31
🤖 Replace Object.fromEntries with Array.reduce

snorkypie:
Rationale behind change:
Object.fromEntries needs to be polyfilled in any project using this right now
and the switch to Array.reduce removes this need. To me it makes sense not
to have dependencies like that. Sentry for the current project I work with is
filled with `Object.fromEntries is not a function` errors.


dmaretskyi:
Thanks for the PR! Just out of curiosity, what platform are you running on? I believed that this was already [implemented](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries) in major browsers.

packages/core/src/providers/chainState/multiChainStates/provider.tsx
```diff
@@ -27,14 +26,15 @@ interface Props {
 }
 
 function composeChainState(networks: Providers, state: State, multicallAddresses: Props['multicallAddresses']) {
-  return fromEntries(
-    Object.keys(networks).map((chainId) => [
-      Number(chainId),
-      {
+  return Object.keys(networks).reduce(
```
dmaretskyi, line 29: instead of removing `fromEntries` helper I would put the Array.reduce logic in there

dmaretskyi:
We might also lower the ES version in the tsconfig if we want to support older platforms

snorkypie:
> Thanks for the PR! Just out of curiosity, what platform are you running on? I believed that this was already [implemented](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries) in major browsers.

Hey! It's a node project. I am not affected personally since I run up-to-date browsers but our Sentry show that enough people are hit with this to require us to do something about it.

snorkypie:
> We might also lower the ES version in the tsconfig if we want to support older platforms

It doesn't really matter what direction is taken for me personally, as long as our project don't have to resort to a slower solution to solve it. What direction are you leaning towards? Keep Array.reduce but leave the fromEntries helper in there?

packages/core/src/providers/chainState/multiChainStates/provider.tsx
snorkypie: The main problem I had with fromEntries to begin is that it feels a bit "hackish" to return an array just to flip it to an object directly after. I personally don't think keeping fromEntries() helps in this case since we'd still have to go the route of creating an array and then flipping it into an object, instead of directly going for the object. Just my 2c.

dmaretskyi: I see your point, something like `mapValues` would be a better abstraction

dmaretskyi:
> It doesn't really matter what direction is taken for me personally, as long as our project don't have to resort to a slower solution to solve it. What direction are you leaning towards? Keep Array.reduce but leave the fromEntries helper in there?

I would leave some sort of helper in there (fromEntries or mapValues). Also I would lower the "lib" version in tsconfig since otherwise it would be difficult for us to enforce those changes.

Do you mind sending the names and versions of affected browsers? Object.fromEntries might not be the only problem in that case.

pawelpolak2:
Thanks for your PR. Looks like it might not be relevant anymore so I'm closing the PR for now. Feel free to reopen if it still affects you. 😸

PR #687, opened 2022-04-13
🎭  Introduce playwright tests for example app

rzadp:


packages/example/playwright/utils/log.ts
```diff
@@ -0,0 +1,5 @@
+import debug from 'debug'
+
+const log = debug('usedapp:example:playwright')
```
dmaretskyi, line 3:
```suggestion
export const log = debug('usedapp:example:playwright')
```

packages/example/playwright/1.without-metamask.test.ts
```diff
@@ -0,0 +1,35 @@
+import { expect } from 'chai'
```
dmaretskyi, line 1: are `1.*` and `2.*` name important for order? can you please add a comment if so

dmaretskyi (approved):

packages/example/playwright/utils/log.ts
rzadp: ✅

PR #695, opened 2022-04-14
⚡️ Improved handling of Falsy parameters in hooks

PoulJohny:
I've found that most of the hooks support passing Falsy as parameters but not all, so I've:
- added possibility to pass Falsy as a contract in useContractFunction
- fixed throwing error when passed Falsy as useToken address parameter


dmaretskyi (approved):

packages/core/src/hooks/useToken.ts
```diff
@@ -8,6 +8,9 @@ import { Contract } from 'ethers'
  * @public
  */
 export function useToken(tokenAddress: string | Falsy): TokenInfo | undefined {
+  if (!tokenAddress) {
```
dmaretskyi, line 11: I don't think you can do that. Every call to the hook must invoke the same set of hooks in the same order every time. This one misses the `useCalls` below.

https://reactjs.org/docs/hooks-rules.html

dmaretskyi (changes requested):

packages/core/src/hooks/useToken.ts
noobshow: I think the result might be undefined regardless of this check.
Didn't test it though.

PoulJohny: @dmaretskyi you are right, knew about it but forgot to change. Now I've pushed updated code. Please check :)

mj426382 (approved):

gpxl-dev:
@dmaretskyi - looks like this PR Is blocked by an outdated change request?
I've hit an issue caused by the error thrown in `useToken` with falsy params and would love to see this merged.

rzadp (approved):

PR #701, opened 2022-04-19
🧠 Deploy multicall2 on localhost

noobshow:
Deploy multicall2 on localhost if the version in config is multicall2


mj426382 (changes requested):
Repair lint in your PR and I think that we can merge :)

mj426382 (approved):

PR #705, opened 2022-04-20
🎨 Linter improvements

yivlad:


packages/docs/src/examples/WalletConnectExample.tsx
```diff
@@ -36,7 +36,7 @@ function App() {
           infuraId: '62687d1a985d4508b2b7a24827551934',
         })
         await provider.enable()
-        activate(provider)
+        await activate(provider)
```
mj426382, line 39: I think that it does not make any difference, but if linter shouts, we can add that.

packages/example/src/pages/WalletConnect.tsx
```diff
@@ -21,7 +21,7 @@ export function WalletConnect() {
         infuraId: 'd8df2cb7844e4a54ab0a782f608749dd',
       })
       await provider.enable()
-      activate(provider)
+      await activate(provider)
```
mj426382, line 24: Same thing here.

mj426382 (approved):

packages/docs/src/examples/WalletConnectExample.tsx
yivlad: Yep, that is the point of `no-floating-promises` rule - you have to either await promise or use the `void` operator to explicitly say that you don't want to await the promise. I think here it makes more sense to await the promise.

PR #716, opened 2022-04-23
Fix CoinGecko Invalid Hook Call Error

javonmcgilberry:
It appears that the coingecko package is incorrectly specifying react as a dependency **rather than a peer dependency**. [Please see here in react docs for more details](https://reactjs.org/warnings/invalid-hook-call-warning.html). 

Under its current form, this results in 2 copies of react. The application breaks with the latest release of React Version 18, because we have hardcoded version 17 (as a workaround the user is now forced to do an npm link)

Interestingly enough, this peer dependency detail is already something we are doing for the core useDApp package, so perhaps this was overlooked by mistake :)


rzadp:
@javonmcgilberry Good catch.
You need to regenerate lockfile - `pnpm install` and commit the changes.

javonmcgilberry:
Sorry about that! @rzadp . I believe we should be good now

mj426382 (commented):
Remove pnpm-lock.yaml and run pnpm install && pnpm build and then push changes.

rzadp (approved):

javonmcgilberry:
alright guys we should be good now 😅

PR #729, opened 2022-04-29
🆙 Add siwe example to docs

mj426382:


rzadp (approved):
Looking good 👍 
One thing - does it make sense to move the doc under `Guides > Connecting` ?

PR #730, opened 2022-05-02
🪱 Attempt adding network if it isn't added to wallet

p-sad:
This is a change to `switchNetwork` that attempts to add a network if it hasn't been added to the extension. As far as I know, both functions only work with Metamask.

In addition to a review I need some opinions: so far I've added connection details to some networks, just to demonstrate my approach and see if it's alright. The documentation probably also needs updating.
Possible next steps:
* Complete details in other known networks
* Add demo of this functionality to the example project


mj426382 (commented):
Looks good:)
Also it is good to have example in docs with that.
WDYT?

mj426382 (approved):

PR #732, opened 2022-05-02
📄 Fill todo docs files

mj426382:


packages/docs/docs/02-Guides/02-Reading/03-Handling Falsy.mdx
```diff
@@ -0,0 +1,31 @@
+# Handling Falsy types
+
+`useDapp` supports handling `False`
+
+```tsx
+export type Falsy = false | 0 | '' | null | undefined
+```
+
+If you pass one of these types to call, it will return undefined without making blockchain query.
+
```
rzadp, line 10:
```suggestion

## Example

```

packages/docs/docs/02-Guides/02-Reading/03-Handling Falsy.mdx
```diff
@@ -0,0 +1,33 @@
+# Handling Falsy types
+
+`useDapp` supports handling `False`
```
rzadp, line 3:
```suggestion
Hooks in `useDapp` support handling "falsy" values in most parameters.
```

packages/docs/docs/02-Guides/02-Reading/03-Handling Falsy.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+```tsx
+export type Falsy = false | 0 | '' | null | undefined
+```
+
+If you pass one of these types to call, it will return undefined without making blockchain query.
```
rzadp, line 9:
```suggestion
If you pass one of these types to a call, it will return `undefined` without making blockchain query. This allows us to conditionally make a call if a parameter is not known ahead of time, without breaking the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html).
```

packages/docs/docs/02-Guides/02-Reading/03-Handling Falsy.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+## Example
+
+```tsx
+export function useTokenBalance(
```
rzadp, line 14: We are interested in an example from the point of view of the user.
So something like:

```
const {account} = useEthers() // account is of type "string | undefined"
const balance = useEtherBalance(account) // Will start making queries once account is defined
```

packages/docs/docs/02-Guides/03-Transactions/04-Replaced Transactions.mdx
```diff
@@ -0,0 +1,10 @@
+# Replaced Transactions
+
+If the status is pending you can replace transaction with increased gas limit to have it mined quicker. To do it in just send another transactions with the same nonce as previous. This transaction is called `replaced transaction`.
+After replacing transaction on wallet, you will see it as a previous transaction(one you replaced) and it will be treated as a previous transaction.
```
rzadp, line 4:
```suggestion
After replacing a transaction on a wallet, Usedapp will detect this and track the new, replaced transaction.
```

packages/docs/docs/02-Guides/03-Transactions/04-Replaced Transactions.mdx
```diff
@@ -0,0 +1,10 @@
...
+
+<img src="https://lh3.googleusercontent.com/FtUnvBTcWuTDZauj0IXI4uY9kB3gxkVHJFsuu5SlTzs0zRt8RNV3GAup9WGrKfw1AoALuoFYM5ZjT320wR2mS3UOozJl8bhisQ2d5Bh0aDqe58x9rzvEYbZbG1OtJm7_HdyBBXs7" alt="replace_metamask"/>
```
rzadp, line 6: Hotlinking is bad - copy the image and put it in our repo.

mj426382: Okay

packages/docs/docs/02-Guides/03-Transactions/04-Replaced Transactions.mdx
```diff
@@ -0,0 +1,10 @@
+# Replaced Transactions
+
+If the status is pending you can replace transaction with increased gas limit to have it mined quicker. To do it in just send another transactions with the same nonce as previous. This transaction is called `replaced transaction`.
```
rzadp, line 3:
```suggestion
If the status is pending you can replace a transaction with increased gas limit to have it mined quicker. To do it just send another transaction with the same nonce as the previous one. The new transaction replaces the previous one.
```

rzadp (approved):

PR #737, opened 2022-05-04
🤡 Add styled examples

mj426382:


packages/docs/src/examples/Siwe.tsx
```diff
@@ -28,6 +28,7 @@ export function App() {
       return (
         <div>
             <button onClick={() => signIn()}>{!isLoggedIn ? 'Sign in' : 'Sign in again'}</button>
+            {' '}
```
rzadp, line 31: Maybe `&nbsp;` instead?

packages/docs/src/examples/Multichain.tsx
```diff
@@ -28,10 +30,16 @@ export function App() {
...
   return (
     <>
-      <div> Account address: {address}</div>
-      <div> Balance on Mainnet: {mainnetBalance && formatEther(mainnetBalance)} Eth </div>
-      <div> Balance on Arbitrum: {arbitrumBalance && formatEther(arbitrumBalance)} AEth </div>
-      <div> Balance on ZkSync Testnet: {zkSyncBalance && formatEther(zkSyncBalance)} ZKEth </div>
+      <div className={styles.Balance}> Account:</div>
+      <div className={styles.Inline}>
+        <AccountIcon account={address}/>
+        {' '}
```
rzadp, line 36: `&nbsp;` instead?

packages/docs/src/examples/ConnectingToNetwork.tsx
```diff
@@ -21,17 +23,54 @@ ReactDOM.render(
...
   const stakingBalance = useEtherBalance(STAKING_CONTRACT)
+
+  const ConnectButton = () => (
+    <div>
+        <button onClick={() => activateBrowserWallet()}>Connect</button>
+        <p>Connect to wallet to interact with the example.</p>
+    </div>
+  )
+
+  const MetamaskConnect = () => (
```
rzadp, line 37: Not using the extracted one?

mj426382: Yes, because i want to show users code single time:)

rzadp: Makes sense.

packages/docs/src/examples/components/MetamaskConnect.tsx
```diff
@@ -0,0 +1,28 @@
...
+        <p>Connect to wallet to interact with the example.</p>
+    </div>)
+
+    return (
+        <div>
+        {account && (
+        <div>
+            <div id="Inline_TYMo">
+                <AccountIcon account={account}/>
+                {' '}
```
rzadp, line 20: nbsp

packages/docs/src/examples/WalletConnectExample.tsx
```diff
@@ -41,13 +42,42 @@ function App() {
...
+      </div>
+    )
+  
+    const WalletConnectConnect = () => (
+      <div>
+        {account && (
+        <div>
+            <div id="Inline_TYMo">
+                <AccountIcon account={account}/>
+                {' '}
```
rzadp, line 58: nbsp

packages/docs/src/examples/ExampleContainer.tsx
```diff
@@ -3,6 +3,7 @@ import CodeWrapper from './CodeWrapper'
 import BrowserOnly from '@docusaurus/BrowserOnly'
 import  ReactDOM from 'react-dom'
 import { ErrorBoundary, FallbackProps } from 'react-error-boundary'
+import './styles/styles.module.css'
```
rzadp, line 6: Is this not imported twice?

packages/docs/src/examples/styles/styles.module.css
```diff
@@ -0,0 +1,19 @@
...
+  font-weight: 450;
+}
+
+#Balance {
+  font-size: 16px;
+  color: grey;
+  display: inline-flexbox;
+}
+
+#Bold {
```
rzadp, line 12: You can have more than 1 bold item on a page, so this should be a class not an ID - same with others.

packages/docs/src/examples/ConnectingToNetwork.tsx
```diff
@@ -21,17 +22,54 @@ ReactDOM.render(
...
+    <div>
+      {account && (
+      <div>
+          <div id="Inline_TYMo">
```
rzadp, line 40: This `_xxxx` looks bad. It looks ugly, and it can change in the future.
We should change webpack to not add it anymore.

packages/docs/src/examples/ExampleContainer.tsx
mj426382: yes. good catch

rzadp (approved):

PR #749, opened 2022-05-09
🌀 Auto generate hooks API documentation

rzadp:
Different version of https://github.com/TrueFiEng/useDApp/pull/739 based on `typedoc` instead of `jsdoc2md`.

Enables auto-generated documentation of hooks from jsdoc comments and TS types.


packages/core/src/hooks/useDebouncePair.ts
```diff
@@ -1,8 +1,15 @@
 import { useEffect, useState } from 'react'
 
-// modified from https://usehooks.com/useDebounce/
 /**
+ * Debounce a pair of values of types T and U.
+ * It stores a single value but returns after debounced time unless a new value is assigned before the debounce time elapses, in which case the process restarts.
+
+ * This function is used for debouncing multicall until enough calls are aggregated.
+ * @param first first variable to be debounced
```
yivlad, line 8:
```suggestion
 * @param first variable to be debounced
```

packages/core/src/hooks/useDebouncePair.ts
```diff
@@ -1,8 +1,15 @@
...
+ * @param second second variable to be debounced
```
yivlad, line 9:
```suggestion
 * @param second variable to be debounced
```

packages/core/src/hooks/useEtherBalance.ts
```diff
@@ -7,7 +7,18 @@ import { useCall } from './useCall'
 import { Contract } from 'ethers'
 
 /**
+ * Returns ether balance of a given account.
+ * @param address address of an account
```
yivlad, line 11:
```suggestion
 * @param address of an account
```

packages/docs/generate/generate.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Go to root of docs package
+cd $(dirname "$0")/..
+
+ts-node -r ts-node/register/transpile-only \
```
yivlad, line 7: Shouldn't this be?
```suggestion
ts-node -r ts-node/register/transpile-only \
```

yivlad (approved):

packages/core/src/hooks/useChainCalls.ts
```diff
@@ -3,17 +3,25 @@ import { Falsy } from '../model/types'
...
+ * @param calls list of calls, also see {@link RawCall}. Calls need to be in the same order across component renders.
+ * @returns encoded result or Falsy value if call didn't return yet or an error occurred.
  */
 export function useChainCalls(calls: (RawCall | Falsy)[]) {
   const results = useRawCalls(calls)
   return results.map((result) => result?.value)
 }
 
 /**
+ * Makes a call to a specific contract and returns the value. The hook will cause the component to refresh whenever a new block is mined and the value is changed.
```
mj426382, line 19:
```suggestion
 * Makes a call to a specific contract and returns its value. The hook will cause the component to refresh whenever a new block is mined and the value is changed.
```

packages/docs/docs/03-API Reference/03-Models.mdx
```diff
@@ -63,6 +63,10 @@ Enables reconnecting to last used provider when user revisits the page.
   }
 ```
 
+## ChainId
+
+A number describing the ID of the chain. See [ChainList](https://chainlist.org/) for a list of known chains and their IDs.
+
```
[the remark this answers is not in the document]
rzadp: Good idea, added.

packages/core/src/hooks/useEthers.ts
```diff
@@ -37,7 +38,19 @@ export type Web3Ethers = {
 }
 
 /**
+ * Returns connection state and functions that allow to manipulate the state.
+ * **Requires**: `<ConfigProvider>`
+ * 
  * @public
+ * @returns {} Object with the following:
```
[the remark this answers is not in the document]
rzadp: Good catch, I will pin it down in the next PR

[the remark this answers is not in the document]
rzadp: https://github.com/TrueFiEng/useDApp/pull/753

PR #752, opened 2022-05-09
🎍 Add example to `getChainName` docs

rzadp:
Closes https://github.com/TrueFiEng/useDApp/issues/751

### DIFF
--- a/packages/docs/docs/03-API Reference/05-Helpers.mdx
+++ b/packages/docs/docs/03-API Reference/05-Helpers.mdx
@@ -97,7 +97,7 @@ Can call with Chain directly:
 > // Mainnet
 ```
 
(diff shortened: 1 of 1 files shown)


packages/docs/docs/03-API Reference/05-Helpers.mdx
```diff
@@ -97,6 +97,14 @@ Can call with Chain directly:
 > // Mainnet
 ```
 
+In order to retrieve the chain name given just the `chainId`, in can be done like so:
```
yivlad, line 100:
```suggestion
In order to retrieve the chain name given just a `chainId`, in can be done like so:
```

yivlad (approved):

PR #768, opened 2022-05-19
✉️ Add useResolveName hook

yivlad:


packages/core/src/hooks/useLookupAddress.ts
```diff
@@ -2,39 +2,45 @@ import { useEffect, useState } from 'react'
...
+    })()
 
     return () => {
       mounted = false
     }
-  }, [account, library])
+  }, [address, library])
 
-  return ens
+  return { ens, isLoading, error }
```
rzadp, line 45: A breaking change, we need a version bump and a migration guide.

rzadp (approved):

PR #775, opened 2022-05-24
🥝 Add fast multicall encoding option

yivlad:


packages/core/src/constants/type/Config.ts
```diff
@@ -32,6 +32,10 @@ export type FullConfig = {
    * Version of multicall contract on the chain.
    */
   multicallVersion: 1 | 2
+  /**
+   * @experimental
+   */
+  fastMulticallEncoding: boolean
```
rzadp, line 38:
```suggestion
  fastMulticallEncoding?: boolean | undefined
```

packages/core/src/model/config/default.ts
```diff
@@ -13,4 +13,5 @@ export const DEFAULT_CONFIG: FullConfig = {
   },
   autoConnect: true,
   multicallVersion: 1,
+  fastMulticallEncoding: false,
```
rzadp, line 16:
```suggestion
```

packages/core/src/abi/common/index.ts
```diff
@@ -0,0 +1,8 @@
+export const buffLength = (buf: string) => (buf.length - 2) / 2
```
rzadp, line 1: Could just be `common.ts` file.

Can you add links or explanation of what's going on in here, as much as possible?

packages/core/src/abi/multicall/encoder.ts
```diff
@@ -0,0 +1,14 @@
+import { encodeUint } from '../common'
+import { encodeCalls } from '../multicall2/encoder'
+import { ethersAbi } from './constants'
+
+const selector = ethersAbi.getSighash('aggregate')
+
+export function encodeAggregate(calls: [string, string][]) {
+  // head params
+  const dynamicOffset = 0x20
+  let res = selector
```
rzadp, line 10: Why not `const res = selector + encodeUint(dynamicOffset)`?

packages/core/src/abi/multicall/encoder.ts
```diff
@@ -0,0 +1,14 @@
+import { encodeUint } from '../common'
+import { encodeCalls } from '../multicall2/encoder'
+import { ethersAbi } from './constants'
+
+const selector = ethersAbi.getSighash('aggregate')
+
+export function encodeAggregate(calls: [string, string][]) {
+  // head params
+  const dynamicOffset = 0x20
```
rzadp, line 9: Where is this number coming from?

packages/core/src/abi/multicall2/decoder.ts
```diff
@@ -0,0 +1,40 @@
...
+  const arraySize = getNumber(1)
+  const calls: [boolean, string][] = []
+
+  for (let i = 0; i < arraySize; i++) {
+    // offset of the call number i
+    const callOffset = 2 * getNumber(i + 2) + 2 * wordLength
+    // position of the call if we split calldata in chunks of 32 bytes
+    const pos = callOffset / wordLength
+    // returnData is encoded as a flag showing if the call was successful,
+    // data offste, which sould be equal to 0x40, data length and the data itself
```
rzadp, line 24:
```suggestion
    // data offset, which sould be equal to 0x40, data length and the data itself
```

rzadp (approved):

PR #780, opened 2022-05-26
🔒 Fix ethers versions

yivlad:


package.json
```diff
@@ -16,12 +16,12 @@
   },
   "dependencies": {
     "@changesets/cli": "^2.14.1",
-    "ethers": "^5.6.2",
+    "ethers": "5.6.2",
```
mj426382, line 19: Why did you 'downgrade' ethers version?

yivlad: To make usedapp predictable. Otherwise it just installs the latest version of ethers compatible with 5.6.2

mj426382 (approved):

PR #784, opened 2022-05-30
❗️ Fix exception during estimating gas price

mj426382:


packages/core/src/hooks/usePromiseTransaction.ts
```diff
@@ -16,10 +16,14 @@ export async function estimateGasLimit(
...
-    ? BigNumber.from(transactionRequest.gasLimit)
-    : await signer.estimateGas(transactionRequest)
-  return estimatedGas?.mul(bufferGasLimitPercentage + 100).div(100)
+  try {
+    const estimatedGas = transactionRequest.gasLimit
+      ? BigNumber.from(transactionRequest.gasLimit)
+      : await signer.estimateGas(transactionRequest)
+    return estimatedGas?.mul(bufferGasLimitPercentage + 100).div(100)
+  } catch (err: any) {
+    return undefined
```
rzadp, line 25: Better not loose it in case it's useful.

```suggestion
    console.error(err)
    return undefined
```

rzadp (commented):
Let's add a changeset

rzadp (approved):

PR #786, opened 2022-05-30
🐊 Disable fetching data while window inactive

mj426382:


packages/core/src/providers/window/provider.tsx
```diff
@@ -0,0 +1,26 @@
...
+  children: ReactNode
+}
+
+/**
+ * @internal Intended for internal use - use it on your own risk
+ */
+export function WindowProvider({ children }: Props) {
+  const [isActiveWindow, setActiveWindow] = useState(true)
+
+  document.addEventListener('visibilitychange', () => {
```
rzadp, line 14: Shouldn't this be in a useEffect?
We have no guarantee that this will be executed only once.

packages/core/src/providers/blockNumber/blockNumber/provider.tsx
```diff
@@ -14,8 +15,11 @@ interface Props {
 export function BlockNumberProvider({ children }: Props) {
   const { library, chainId } = useEthers()
   const [state, dispatch] = useReducer(blockNumberReducer, {})
+  const { isActive } = useWindow()
 
-  useEffect(() => subscribeToNewBlock(library, chainId, dispatch), [library, chainId])
+  useEffect(() => {
+    subscribeToNewBlock(library, chainId, dispatch, isActive)
```
rzadp, line 21: This changed the behaviour!

```suggestion
    return subscribeToNewBlock(library, chainId, dispatch, isActive)
```

rzadp (commented):
Can you make sure to test it in chrome, firefox and safari?

packages/core/src/providers/window/provider.tsx
mj426382: It should be, thanks for catch:)

packages/core/src/providers/blockNumber/blockNumber/provider.tsx
mj426382: My mistake with adding these brackets

packages/core/src/providers/window/provider.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+/**
+ * @internal Intended for internal use - use it on your own risk
+ */
+export function WindowProvider({ children }: Props) {
+  const [isActiveWindow, setActiveWindow] = useState(true)
+
+  useEffect(() => {
+    document.addEventListener('visibilitychange', () => {
```
rzadp, line 15: We should add an event listener, and remove on unmounting.

So at the end of useEffect
```
return () => document.removeEventListener(...)
```

rzadp (approved):

PR #795, opened 2022-06-02
⚡️ Add block refresher

mj426382:


packages/core/src/constants/type/Config.ts
```diff
@@ -71,6 +71,10 @@ export type FullConfig = {
    * Enables reconnecting to last used provider when user revisits the page.
    */
   autoConnect: boolean
+  /**
+   * Refresh standard calls each time the n-th block is mined.
+   */
+  refresh?: number | string
```
rzadp, line 77: Specify available strings exactly, not string.

rzadp (changes requested):

packages/core/src/hooks/useCall.ts
```diff
@@ -76,10 +77,15 @@ export function useCall<T extends TypedContract, MN extends ContractMethodNames<
...
   const chainId = useChainId({ queryParams })
-  const { isStatic } = queryParams
+  const { refresh } = useConfig()
 
   const rawCalls = useMemo(
-    () => calls.map((call) => (chainId !== undefined ? encodeCallData(call, chainId, isStatic) : undefined)),
+    () =>
+      calls.map((call) =>
+        chainId !== undefined
+          ? encodeCallData(call, chainId, { ...queryParams, refresh: queryParams.refresh ?? refresh })
```
rzadp, line 86: Can it be simplified into `{ refresh, ...queryParams }`?

mj426382: Yes, it is leftover after rename refreshPerBlocks to refresh

packages/core/src/constants/type/QueryParams.ts
```diff
@@ -12,6 +12,7 @@ export interface QueryParams {
    */
   chainId?: ChainId
   isStatic?: boolean
+  refresh?: number | 'never' | 'everyBlock'
```
rzadp, line 15:
```suggestion
  refresh?: Config['refresh']
```

We can tie the types into one place.

packages/core/src/providers/chainState/multiChainStates/provider.tsx
```diff
@@ -82,6 +82,7 @@ export function MultiChainStateProvider({ children, multicallAddresses }: Props)
 
     const callsOnThisChain = uniqueCalls.filter((call) => call.chainId === chainId)
     if (callsOnThisChain.length === 0) {
+      dispatchCalls({ type: 'UPDATE_CALLS', calls, blockNumber, chainId })
```
rzadp, line 85: Comment why this is needed here?

packages/core/src/hooks/useCall.ts
rzadp: Yes, then queryParams will override.

packages/core/src/providers/chainState/multiChainStates/provider.tsx
```diff
@@ -81,9 +81,7 @@ export function MultiChainStateProvider({ children, multicallAddresses }: Props)
     }
 
     const callsOnThisChain = uniqueCalls.filter((call) => call.chainId === chainId)
-    if (callsOnThisChain.length === 0) {
```
rzadp, line 84: Why do we remove this?

mj426382: i put that on multicall function, to reduce duplicated code

PR #798, opened 2022-06-03
👯 Ethers as peer dep

yivlad:


mj426382 (approved):

packages/coingecko/package.json
```diff
@@ -36,13 +36,12 @@
...
   "dependencies": {
-    "@ethersproject/contracts": "5.6.0",
-    "@ethersproject/providers": "5.6.2",
     "@usedapp/core": "workspace:*",
     "isomorphic-fetch": "^3.0.0"
   },
   "peerDependencies": {
-    "react": "*"
+    "react": "*",
+    "ethers": "*"
```
rzadp, line 44: Don't we need ethers in dev dependencies? Or maybe we don't need ethers here at all?

packages/core/package.json
```diff
@@ -10,34 +10,24 @@
...
     "react-flat-providers": "^2.1.0"
   },
   "peerDependencies": {
-    "react": "*"
+    "react": "*",
+    "ethers": "*"
   },
   "devDependencies": {
     "@ethereum-waffle/provider": "4.0.0-alpha.0",
+    "@ethersproject/abstract-provider": "^5.6.1",
```
rzadp, line 30: Is it not included in ethers?

packages/siwe/package.json
```diff
@@ -20,7 +20,8 @@
   ],
   "peerDependencies": {
     "@usedapp/core": "*",
-    "react": "*"
+    "react": "*",
+    "ethers": "*"
```
rzadp, line 24: is it actually needed? don't we need it in dev dependency?

packages/coingecko/package.json
yivlad: This one is needed in case someone installs `@usedapp/coingecko` and not `@usedapp/core`. Because peer dependencies are not transitive, we need to include `ethers` in peer dependencies here as well.

packages/core/package.json
yivlad: Not everything is exported in a nice way. Anyway we need only types from `@ethersproject/abstract-provider`, so I think this is ok.

rzadp (approved):

PR #802, opened 2022-06-03
🧺 Add refreshing docs

mj426382:


packages/docs/docs/02-Guides/02-Reading/04-Refresh Per Blocks.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+```tsx
+  import { Mainnet } from '@usedapp/core'
+
+  const config = {
+    readOnlyChainId: Mainnet.chainID,
+    readOnlyUrls: {
+      [Mainnet.chainID]: 'https://mainnet.infura.io/v3/62687d1a985d4508b2b7a24827551934',
+    },
+    refresh: 'never',
```
yivlad, line 17: I suppose the common practice in such cases is to set parameter to a negative value, such as -1.

yivlad (approved):

PR #823, opened 2022-06-10
🛵 Track non static calls for each chain

yivlad:


packages/core/src/providers/blockNumber/blockNumbers/provider.tsx
```diff
@@ -16,14 +16,28 @@ export function BlockNumbersProvider({ children }: Props) {
...
+
     const onUnmount = Object.entries(networks).map(([chainId, provider]) =>
       subscribeToNewBlock(provider, Number(chainId), dispatch, isActive)
     )
 
     return () => {
       onUnmount.forEach((fn) => fn())
     }
-  }, [networks, isActive])
+  }, [networks])
```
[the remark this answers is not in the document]
yivlad: The logic was moved to readonlyNetworks provider

mj426382 (approved):
Remake CI and we can merge:)

packages/core/src/providers/chainState/common/chainStateReducer.test.ts
```diff
@@ -37,7 +37,8 @@ describe('chainStateReducer', () => {
     expect(result).to.deep.equal(state)
   })
 
-  it('overwrites with updates from newer blocks', () => {
+  // skip because the logic has changed
+  it.skip('overwrites with updates from newer blocks', () => {
```
rzadp, line 41: Can the test be updated with the new logic, to make sure the state is changed in a predictable way?

yivlad: The new logic is the same as in case of merging the same block, so we already have a test for it

rzadp: But the setup is different, if the outcome is the same it can be copied to this test as well

packages/core/src/hooks/useBlockMeta.ts
```diff
@@ -14,13 +15,31 @@ const GET_CURRENT_BLOCK_DIFFICULTY_CALL = MultiCallABI.encodeFunctionData('getCu
  */
 export function useBlockMeta(queryParams: QueryParams = {}) {
   const chainId = useChainId({ queryParams })
+  const { refresh: configRefresh } = useConfig()
```
rzadp, line 18: Hmm can you help me understand why `useBlockMeta` is the only one hook that had to be changed in this PR?

mj426382 (approved):

rzadp (approved):

PR #845, opened 2022-06-24
👨🏻‍💼 Add private key sending transactions

mj426382:
closes https://github.com/TrueFiEng/useDApp/issues/702


packages/core/src/model/TransactionOptions.ts
```diff
@@ -6,4 +6,9 @@ import { Signer } from 'ethers'
 export interface TransactionOptions {
   signer?: Signer
   transactionName?: string
+  privateKey?: string
+  mnemonicPhrase?: string
+  encryptedJson?: string
+  chainId?: number // privateKey, mnemonicPhrase and encryptedJso
```
rzadp, line 12: I don't understand the comment?

packages/core/src/model/TransactionOptions.ts
```diff
@@ -6,4 +6,9 @@ import { Signer } from 'ethers'
...
+  password?: string // for encryptedJson purposes
```
rzadp, line 13: Hmm, how about something like:

```ts
encryptedJson?: { json: string, password: string }
```

packages/docs/docs/02-Guides/03-Transactions/08-Sendign Without Wallet.mdx
```diff
@@ -0,0 +1,49 @@
+# Sending without Wallet
```
rzadp, line 1: Typo in file name.

packages/core/src/hooks/useSendTransaction.test.ts
```diff
@@ -51,22 +64,81 @@ describe('useSendTransaction', () => {
     expect(result.current.state.errorMessage).to.eq('invalid address')
   })
 
-  it('Returns receipt after correct transaction', async () => {
```
rzadp, line 54: Shouldn't we keep this test as well?

packages/core/src/model/TransactionOptions.ts
mj426382: Solved :)

packages/core/src/model/TransactionOptions.ts
```diff
@@ -6,4 +6,9 @@ import { Signer } from 'ethers'
 export interface TransactionOptions {
   signer?: Signer
   transactionName?: string
+  privateKey?: string
```
yivlad, line 9: Does signer and privateKey fields make sense being placed in the same object? If not maybe something like this:
```ts
export interface TransactionOptionBase {
  transactionName?: string
}
export interface TransactionOptionWithSigner {
  signer?: Signer
}
export interface TransactionOptionsWithPrivateKey {
  privateKey?: string,
  chainId?: number
}
...
export type TransactionOptions = TransactionOptionBase & (TransactionOptionWithSigner | TransactionOptionsWithPrivateKey | ...)
```

packages/core/src/helpers/getSignerFromOptions.ts
```diff
@@ -0,0 +1,26 @@
+import { ethers, providers } from 'ethers'
```
mj426382, line 1: Add some test for this

yivlad (approved):

PR #848, opened 2022-06-27
📀 Add own local storage implementation

mj426382:
closes https://github.com/TrueFiEng/useDApp/issues/846


packages/core/src/constants/type/Config.ts
```diff
@@ -88,6 +93,10 @@ export type FullConfig = {
    * Refresh standard calls each time the n-th block is mined.
    */
   refresh?: number | 'never' | 'everyBlock'
+  /**
+   * Local storage overriding purposes
```
rzadp, line 97:
```suggestion
   * Optional Local storage override for use in environments like React Native
```

packages/core/src/constants/type/Config.ts
```diff
@@ -88,6 +93,10 @@ export type FullConfig = {
...
+   */
+  localStorageFunctions: LocalStorageFunctions
```
rzadp, line 99: Consider using `WindowLocalStorage['localStorage']` type instead of creating your own.

rzadp, line 99:
```suggestion
  localStorageOverride: LocalStorageFunctions
```

packages/core/src/hooks/useLocalStorage.ts
```diff
@@ -1,34 +1,13 @@
...
-  }
-}
+import { useConfig } from './useConfig'
 
 /**
  * @internal Intended for internal use - use it on your own risk
  */
 export function useLocalStorage(key: string) {
+  const {
+    localStorageFunctions: { getItem, setItem },
```
rzadp, line 9: There might be a problem with this  destructuring approach, if someone provides a class which relies on `this` for keeping the keys.

packages/core/src/hooks/useLocalStorage.ts
```diff
@@ -1,11 +1,12 @@
 import { useEffect, useState } from 'react'
+import { useConfig } from './useConfig'
 
-function getItem(key: string) {
+function getItem(key: string, localStorage: WindowLocalStorage['localStorage']) {
   if (typeof window === 'undefined') {
```
yivlad, line 5: I don't think this is necessary anymore

packages/core/src/constants/type/Config.ts
```diff
@@ -88,6 +88,10 @@ export type FullConfig = {
    * Refresh standard calls each time the n-th block is mined.
    */
   refresh?: number | 'never' | 'everyBlock'
+  /**
+   * Optional Local storage override for use in environments like React Native
+   */
+  localStorageOverride: WindowLocalStorage['localStorage']
```
rzadp, line 94:
```suggestion
  localStorageOverride?: WindowLocalStorage['localStorage']
```

The average user should not have to set it, or know that this exists.

packages/core/src/hooks/useLocalStorage.ts
mj426382: I think that it is, because we create own functions from localStorage implementation.

packages/core/src/hooks/useLocalStorage.ts
```diff
@@ -29,17 +30,20 @@ function setItem(key: string, value: any) {
  * @internal Intended for internal use - use it on your own risk
  */
 export function useLocalStorage(key: string) {
-  const [value, setValue] = useState(() => getItem(key))
+  const {
+    localStorageOverride = typeof window !== 'undefined' ? window.localStorage : global.localStorage,
+  } = useConfig()
+  const [value, setValue] = useState(getItem(key, localStorageOverride))
```
rzadp, line 36:
```suggestion
  const [value, setValue] = useState(() => getItem(key, localStorageOverride))
```

Preserve previous behaviour.

packages/core/src/hooks/useLocalStorage.ts
```diff
@@ -1,11 +1,12 @@
 import { useEffect, useState } from 'react'
+import { useConfig } from './useConfig'
 
-function getItem(key: string) {
+function getItem(key: string, storage: WindowLocalStorage['localStorage']) {
   if (typeof window === 'undefined') {
```
rzadp, line 5: Not gonna work even with override, because of this.

Add a red test for useLocalStorage with override.

rzadp (approved):

PR #849, opened 2022-06-27
🫧 Multicall fail return error in useCall

yivlad:


packages/core/src/hooks/useRawCalls.ts
```diff
@@ -55,5 +56,19 @@ export function useRawCall(call: RawCall | Falsy) {
...
 function extractCallResult(chains: MultiChainState, call: RawCall): RawCallResult {
   const chainId = call.chainId
-  return chainId !== undefined ? chains[chainId]?.value?.state?.[call.address.toLowerCase()]?.[call.data] : undefined
+  if (chainId !== undefined) {
+    const rawCallResult = chains[chainId]?.value?.state?.[call.address.toLowerCase()]?.[call.data]
+    if (rawCallResult) {
+      return rawCallResult
+    }
+    const error = chains[chainId]?.value?.error as any
+    const errorMessage = error?.error?.message ?? error?.message
```
rzadp, line 65: I hope we don't need to add more, like here:

https://github.com/TrueFiEng/useDApp/blob/c074e166e1f80b5567d8c002ec06dcb16abcfe8f/packages/core/src/hooks/usePromiseTransaction.ts#L84

packages/core/src/hooks/useCall.test.tsx
```diff
@@ -26,163 +26,198 @@ describe('useCall', () => {
...
+            },
+            multicallVersion,
+          }
+        )
+        await waitForCurrent((val) => val !== undefined)
+        expect(result.error).to.be.undefined
+        expect(result.current?.value[0]).to.eq(MOCK_TOKEN_INITIAL_BALANCE)
+      })
+
+      it('initial test balance to be correct', async () => {
```
rzadp, line 62: Copy-pasted test name

packages/core/src/hooks/useCall.test.tsx
```diff
@@ -26,163 +26,198 @@ describe('useCall', () => {
...
+          {
+            readonlyMockProviders: {
+              [chainId]: mockProvider,
+            },
+            multicallVersion,
+          }
+        )
+        await waitForCurrent((val) => val !== undefined)
+        expect(result.current?.value).to.be.undefined
+        expect(result.current?.error).to.not.be.undefined
```
rzadp, line 79: Can we assert anything about the error?

mj426382 (approved):

PR #855, opened 2022-06-28
🐸 Added error hash returned to transaction status

suresh174:
we can decode this hash and can be used to show error messages. This will be used mostly in custom error messages.


suresh174 (commented):
Added error hash

suresh174 (commented):
lint issue fixed

rzadp (approved):

PR #858, opened 2022-06-30
😶‍🌫️ Add gas limiter to use contract function

mj426382:
closes https://github.com/TrueFiEng/useDApp/issues/713


packages/core/src/hooks/useSendTransaction.test.ts
```diff
@@ -64,6 +64,53 @@ describe('useSendTransaction', () => {
     expect(result.current.state.errorMessage).to.eq('invalid address')
   })
 
+  it('transfer ether with limit', async () => {
```
rzadp, line 67: How are those tests testing the gas limit percentage if we're overriding the `gasLimit` to zero?

mj426382: We override gas limit to 200% of the estimated one.

rzadp (approved):

PR #862, opened 2022-07-06
🍅 Add more pages and playwright tests for next.js example

jakvbs:


packages/example-next/tsconfig.json
```diff
@@ -1,20 +1,31 @@
...
-  "exclude": ["node_modules"]
-}
+  "include": [
+    "next-env.d.ts",
+    "**/*.ts",
+    "**/*.tsx"
+  ],
+  "exclude": [
+    "node_modules",
+    "playwright"
```
rzadp, line 29: Try to avoid mixing linting changes with meaningful changes - it's harder to review.
Do not use arbitraty auto-lints - the only auto-linter should be the one we have in the repo with `lint --fix`

rzadp (changes requested):
Every time you switch a page, the `<Providers>` is rendered from scratch? So you lose all state?

We'd like to have just one entrypoint with just 1 `DappProvider`, and the routing happening inside of this provider.

jakvbs:
Now everything should be fine.

packages/example-next/pages/index.tsx
```diff
@@ -1,29 +1,9 @@
...
-  noMetamaskDeactivate: true,
-}
 
 const Home: NextPage = () => {
-  return (
-    <DAppProvider config={config}>
-      <Balance />
-    </DAppProvider>
-  )
+  return <Balance />
```
rzadp, line 6: Do we need this? We already have a balance page?

rzadp (commented):
- Let's merge `Providers` and `Usedapp` files.

rzadp (approved):

PR #873, opened 2022-07-12
🐨 Export Falsy and Params Types from core

jakvbs:


packages/core/src/index.ts
```diff
@@ -72,6 +72,7 @@ export {
   useResolveName,
 } from './hooks'
 export type { CurrencyFormatOptions, TransactionOptions, TransactionState, TransactionStatus } from './model'
+export type { Falsy, Params } from './model'
```
rzadp, line 75: join with export above

rzadp (approved):

PR #885, opened 2022-07-15
♨️ Properly update error when switching to not configured network

jakvbs:
Closes https://github.com/TrueFiEng/useDApp/issues/533

### DIFF
--- a/packages/core/src/hooks/useEthers.test.tsx
+++ b/packages/core/src/hooks/useEthers.test.tsx
@@ -2,8 +2,14 @@ import { expect } from 'chai'
 import { Wallet } from 'ethers'
 import { useEffect } from 'react'
 import { Config } from '../constants'
-import { Mainnet } from '../model'
-import { renderDAppHook, SECOND_TEST_CHAIN_ID, setupTestingConfig, TestingNetwork } from '../testing'
+import { Localhost, Mainnet } from '../model'
+import {
+  createMockProvider,
+  renderDAppHook,
+  SECOND_TEST_CHAIN_ID,
+  setupTestingConfig,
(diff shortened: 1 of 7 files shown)


packages/core/src/hooks/useEthers.ts
```diff
@@ -64,14 +65,25 @@ export function useEthers(): Web3Ethers {
     isLoading,
   } = useNetwork()
 
-  const { networks } = useConfig()
+  const { networks, readOnlyUrls } = useConfig()
+  const error = useRef<Error | undefined>()
```
rzadp, line 69: Why `useRef`? Is it needed? Why not just `useState`?

jakvbs: Ah yes, I forgot to change it

packages/core/src/hooks/useEthers.test.tsx
```diff
@@ -58,6 +64,25 @@ describe('useEthers', () => {
...
+        }, [])
+
+        return useEthers()
+      },
+      { config }
+    )
+
+    await waitForCurrent((val) => !!val.error)
+    expect(result.current.error).not.to.be.undefined
+    expect(result.current.error?.toString()).to.include(`Not configured chain id: ${Localhost.chainId}`)
```
rzadp, line 83: Can we have test for both not configured and not supported chains?

jakvbs: There are different error messages for both cases, but we can only check if the error occured. Should we?

rzadp: I meant 2 tests

PR #890, opened 2022-07-18
🔆 Prevent merging with mixed changesets

jakvbs:


.github/workflows/CI.yml
```diff
@@ -92,3 +92,21 @@ jobs:
...
+      - run: npm i -g pnpm@7.1.9
+      - run: pnpm install --frozen-lockfile
+      - run: pnpm build
+      - name: Changesets publish dry run
+        run: |
+          set -e
+          pnpm changeset publish --dry-run
+          pnpm changeset version
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_BOT_TOKEN }}
```
rzadp, line 112: Do we need this?

.github/workflows/CI.yml
```diff
@@ -92,3 +92,21 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '16.x'
+      - run: npm i -g pnpm@7.1.9
+      - run: pnpm install --frozen-lockfile
+      - run: pnpm build
```
rzadp, line 105: Do we actually need to build it?

.github/workflows/CI.yml
```diff
@@ -92,3 +92,14 @@ jobs:
...
+  check-mixed-changesets:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '16.x'
+      - run: npm i -g pnpm@7.1.9
+      - run: pnpm install --frozen-lockfile
```
rzadp, line 104:
```suggestion
      - run: pnpm install --frozen-lockfile --filter="usedapp"
```
So that it is faster

PR #891, opened 2022-07-19
🫠 Add playwright test for siwe in docs

jakvbs:


packages/docs/playwright/with-metamask.test.ts
```diff
@@ -62,20 +62,57 @@ describe(`Browser: ${browserType.name()} with Metamask`, () => {
 
   describe('Guides/Transactions', () => {
     it('Switches networks', async () => {
+      let popupPromise
```
rzadp, line 65: Is this needed?

PR #902, opened 2022-07-26
👩‍🔬 Add throw error on 'switchNetwork'

phanius:
Throw error on ```switchNetwork``` to create options to handle error.

e.g. user reject switch network, so we can disconnect wallet.


phanius:
HI @rzadp I'm not sure why the test fail. it doesn't seem to relate to this change. Also, I have run test on my local and it has pass. Could you have a look into this please. Thank you
image

rzadp:
@phanius This is really weird, all tests pass on my machine as well..

phanius:
@rzadp Oh, It pass now lol. Thank you

rzadp (approved):

PR #909, opened 2022-07-31
⛽ Fix gas limit override in useContractFunction hook

dimlbc:
**Description**

This fix adds opportunity to override gasLimit property when passing it to `send` method of useContractFunction hook.

It can be useful for troubleshooting and debugging failed transactions. Passing gasLimit helps to push transaction to the wallet even if transaction is going to fail.

### DIFF
--- a/packages/core/src/hooks/useContractFunction.ts
+++ b/packages/core/src/hooks/useContractFunction.ts
@@ -99,8 +99,8 @@ export function useContractFunction<T extends TypedContract, FN extends Contract
         )
 
         const modifiedOpts = {
-          ...opts,
           gasLimit,
+          ...opts,
         }
         const modifiedArgs = hasOpts ? args.slice(0, args.length - 1) : args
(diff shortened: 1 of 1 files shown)


D-Ig:
> Passing gasLimit helps to push transaction to the wallet even if transaction is going to fail.

just leaving here a couple of links to prove this statement in case anyone's intrested:

* https://github.com/ethers-io/ethers.js/issues/903#issuecomment-648398488
* https://docs.ethers.io/v5/troubleshooting/errors/#help-UNPREDICTABLE_GAS_LIMIT:~:text=In%20almost%20all%20cases%2C%20this%20will%20unfortunately%20require%20you%20specify%20an%20explicit%20gasLimit%20for%20your%20transaction%2C%20which%20will%20disable%20ether%27s%20automatic%20population%20of%20the%20gasLimit%20field%2C%20which%20will%20cause%20this%20error%20to%20go%20away.

yivlad (approved):

PR #919, opened 2022-08-04
📃 Add missing doc about gasLimitBufferPercentage in TransactionOptions

rzadp:
Progresses https://github.com/TrueFiEng/useDApp/issues/900

### DIFF
--- a/packages/docs/docs/03-API Reference/03-Models.mdx
+++ b/packages/docs/docs/03-API Reference/03-Models.mdx
@@ -274,7 +274,7 @@ Fields:
 
 - ``transactionName?: string`` - specifies a transaction name. Used by notifications and history hooks.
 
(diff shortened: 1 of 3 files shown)


packages/docs/docs/03-API Reference/03-Models.mdx
```diff
@@ -274,6 +274,16 @@ Fields:
 
 - ``transactionName?: string`` - specifies a transaction name. Used by notifications and history hooks.
 
+- ``gasLimitBufferPercentage?: number`` - If set, adds an additional buffer of gas limit to estimated gas limit before sending a transaction.
```
yivlad, line 277:
```suggestion
- ``gasLimitBufferPercentage?: number`` - If set, adds an additional buffer to estimated gas limit before sending a transaction.
```

yivlad (approved):

PR #922, opened 2022-08-08
🧀 Remove duplicate calls for same data

rzadp:


packages/core/src/hooks/useCallResilency.test.tsx
```diff
@@ -199,6 +202,51 @@ describe('useCall Resilency tests', () => {
...
+            ;(config.readOnlyUrls![1337] as any).call = async function (...args: any[]): Promise<any> {
+              if (args[1] === 2) {
+                // In this test, let's take a look at calls made for blockNumber 2.
+                calls.push(JSON.stringify(args))
+              }
+              return await originalCall.apply(config.readOnlyUrls![1337], args as any)
+            }
+
+            await miners[0].sendTransaction({ to: constants.AddressZero })
+            rerender()
```
yivlad, line 241: Is it necessary?

yivlad (approved):

packages/core/src/hooks/useCallResilency.test.tsx
rzadp: Somehow, yes - without the fix, the test is not red without `rerender`

PR #932, opened 2022-08-15
🪡 Fix example hook

MichaelZalla:
The `useTokenBalance()` hook in the website's example code references a deprecated hook, `useContractCall()`:

```ts
function useTokenBalance(
	tokenAddress: string|Falsy,
	address: string|Falsy): BigNumber|undefined
{
	const [tokenBalance] = useContractCall (
		address && tokenAddress && {
			abi: ERC20Interface, // ABI interface of the called contract
			address: tokenAddress, // On-chain address of the deployed contract
			method: "balanceOf", // Method to be called
			args: [address], // Method arguments - address to be checked for balance
			}
		) ?? [ ];

	return tokenBalance;

}
```

https://github.com/TrueFiEng/useDApp/blob/master/packages/website/src/index.html#L336

This call yields a deprecation warning in editors like VS Code:

image

This PR updates the `useTokenBalance()` implementation to instead call `useCall()`, following the hook's `@example` code as closely as possible:

```ts
function useTokenBalance(
	tokenAddress: string|Falsy,
	address: string|Falsy): BigNumber|undefined
{
	const { value, error } = useCall(
		address && tokenAddress && {
			contract: new Contract(tokenAddress, ERC20Interface), // Contract instance
			method: "balanceOf", // Method to be called
			args: [address], // Method arguments - address to be checked for balance
			}
		) ?? {};

	if(error) {
		console.error(error.message)
		return undefined
	}

	const { balance } = value || {}

	return balance

}
```
Screenshot of the updated page example running on localhost:

image

### DIFF
--- a/packages/website/src/index.html
+++ b/packages/website/src/index.html
@@ -328,22 +328,31 @@
               <div class="window">
                 <pre
                   class="window-pre"
-                ><code class="code" data-lines="1"><span class="code--blue">function</span> useTokenBalance (
+                ><code class="code" data-lines="1"><span class="code--blue">function</span> useTokenBalance(
   tokenAddress: string | Falsy,
   address: string | Falsy
-) {
-  <span class="code--red">const</span> [tokenBalance] =
-    useContractCall (
+): BigNumber | undefined {
+
+  <span class="code--red">const</span> { value, error } =
+    useCall(
       address && tokenAddress && {
-          abi: ERC20Interface,<span class="code--gray"> // ABI interface of the called contract</span>
-          address: tokenAddress,<span class="code--gray"> // On-chain address of the deployed contract</span>
+          contract: new Contract(tokenAddress, ERC20Interface),<span class="code--gray"> // Contract instance</span>
           method: "balanceOf",<span class="code--gray"> // Method to be called</span>
           args: [address],<span class="code--gray"> // Method arguments - address to be checked for balance</span>
         }
-    ) ?? [ ];
- <span class="code--red">return</span> tokenBalance;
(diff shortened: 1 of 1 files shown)


packages/website/src/index.html
```diff
@@ -328,22 +328,31 @@ <h2 class="title text--left">Example</h2>
...
-          args: [address],<span class="code--gray"> // Method arguments - address to be checked for balance</span> 
+          contract: new Contract(tokenAddress, ERC20Interface),<span class="code--gray"> // Contract instance</span>
+          method: "balanceOf",<span class="code--gray"> // Method to be called</span>
+          args: [address],<span class="code--gray"> // Method arguments - address to be checked for balance</span>
         }
-    ) ?? [ ];
- <span class="code--red">return</span> tokenBalance;
+    ) ?? { };
+
+  if(error) {
```
rzadp, line 345:
```suggestion
  if (error) {
```

packages/website/src/index.html
```diff
@@ -328,22 +328,31 @@ <h2 class="title text--left">Example</h2>
...
+    console.error(error.message)
+    <span class="code--red">return</span> undefined
+  }
+
+  <span class="code--red">const</span> { tokenBalance } = value || { }
+
+  <span class="code--red">return</span> tokenBalance;
```
rzadp, line 352:
```suggestion
  <span class="code--red">const</span> value?.[0]
```

rzadp:
@MichaelZalla Thanks! I simplified the example code a little bit.

rzadp (approved):

PR #937, opened 2022-08-19
📼 Add SIWE-GnosisSafe integration

jakvbs:


.github/workflows/CI.yml
```diff
@@ -90,8 +90,17 @@ jobs:
           xvfb-run --auto-servernum --server-args="-screen 0 1280x960x24" -- yarn test:playwright
         env:
           MAINNET_URL: ${{ secrets.MAINNET_PROVIDER_URL }}
+          DOCS_GNOSIS_OWNER_FIRST: ${{ secrets.DOCS_GNOSIS_OWNER_FIRST }}
+          DOCS_GNOSIS_OWNER_SECOND: ${{ secrets.DOCS_GNOSIS_OWNER_SECOND }}
           DEBUG: usedapp*
         working-directory: packages/docs
+      # For CI debugging purposes
```
rzadp, line 97: Remove?

packages/docs/playwright/with-metamask.test.ts
```diff
@@ -8,9 +8,19 @@ import {
   addPageDiagnostics,
   MetaMask,
   metamaskChromeArgs as args,
-  waitForPopup
+  waitForPopup,
```
rzadp, line 11: Was this not linted? See if it can be linted automatically.

packages/docs/playwright/with-metamask.test.ts
```diff
@@ -8,9 +8,19 @@ import {
...
+import {
+  connectToMetamask,
+  firstSign,
+  GNOSIS_SAFE_URL,
+  initGnosisSafe,
+  secondSign,
+  connectToWalletConnect,
+} from './gnosisSafeUtils'
+
+waitForExpect.defaults.timeout = 90000
```
rzadp, line 23: Move to `test-setup.ts` (add if not exists), it's weird to change a default in a random test file because it affects or other test files.

packages/docs/playwright/with-metamask.test.ts
```diff
@@ -68,48 +78,237 @@ describe(`Browser: ${browserType.name()} with Metamask`, () => {
...
+    if (page) await page.close()
+    if (context) await context.close()
+
+    context = await browserType.launchPersistentContext('', {
+      headless: false, // Extensions only work in Chrome / Chromium in non-headless mode.
+      slowMo: 500,
+      args,
+      // For CI debugging purposes
+      // recordVideo: {
+      //   dir: 'playwright/recordings/',
```
rzadp, line 141: remove?

packages/docs/playwright/with-metamask.test.ts
```diff
@@ -68,48 +78,237 @@ describe(`Browser: ${browserType.name()} with Metamask`, () => {
...
+    page = await context.newPage()
+    gnosisSiwePage = await context.newPage()
+    addPageDiagnostics(page)
+  }
+
+  before(() => resetBrowserContext())
+  after(() => context?.close())
+
+  before(async () => {
+    await import('dotenv/config')
```
rzadp, line 162: What is it doing here? We don't have a `.env` file.

packages/docs/src/css/custom.css
```diff
@@ -37,3 +37,7 @@
 [data-theme='dark'] .docusaurus-highlight-code-line {
   background-color: rgba(0, 0, 0, 0.3);
 }
+
+p {
+  margin: 0;
```
rzadp, line 42: Have you checked all pages to see if there are no weird rendering?

jakvbs: Yes, looks good

packages/docs/playwright/with-metamask.test.ts
```diff
@@ -68,48 +76,233 @@ describe(`Browser: ${browserType.name()} with Metamask`, () => {
...
       await page.click(XPath.text('button', 'Sign out'))
-
       await waitForExpect(async () => {
         expect(await page.isVisible(`//*[text()='Not logged in']`)).to.be.true
       })
     })
   })
 })
+
+if (process.env.CI) {
```
rzadp, line 111: How about, instead of this, having:

```
describe(`Browser: ${browserType.name()} with Gnosis Safe`, function () {
  if (!process.env.DOCS_GNOSIS_OWNER_FIRST) this.skip()
```

(If that works).

So that we can see the tests which are skipped.

rzadp (approved):

packages/siwe/src/provider.tsx
```diff
@@ -64,28 +146,67 @@ export const SiweProvider = ({ children, backendUrl, api }: SiweProviderProps) =
...
 
-      void getAuth().then((res) => (res.loggedIn ? setLoggedIn(true) : undefined))
+      try {
+        await signIn({ signature, message })
+      } catch (err) {
+        return setError(err)
+      }
+
+      if (signature === '0x') {
+        localStorage.setItem('siweMessage', JSON.stringify(message))
```
msieczko, line 166: I think it is not safe to store siweMessage in localstorage here. Maybe backend should give a temporary cookie and keep the secret for us?

packages/siwe/src/provider.tsx
```diff
@@ -64,28 +146,67 @@ export const SiweProvider = ({ children, backendUrl, api }: SiweProviderProps) =
...
+    if (!getMessageHash) {
+      getMessageHash = await gnosisSafeContract.getMessageHash(utils.hashMessage(message.prepareMessage()))
+      localStorage.setItem('getMessageHash', getMessageHash as string)
+    }
 
-    await signOut()
-    setLoggedIn(false)
-  }, [setLoggedIn, account, chainId, signOut])
+    const onMultiSigSigned = async () => {
+      gnosisSafeContract.removeListener(gnosisSafeContract.filters.SignMsg(getMessageHash), onMultiSigSigned)
```
msieczko, line 194: Check if `gnosisSafeContract.filters.SignMsg(getMessageHash)` should not be extracted here. Also doesn't `once` remove listener automatically?

msieczko (commented):
I'd bump `siwe` version in dependencies to match `fastify-siwe`.

PR #941, opened 2022-08-22
⚱️ Change Infura id in WalletConnect page

jakvbs:


.changeset/shiny-gifts-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+"docs": patch
+---
+
+Change Infura id in WalletConnect page
```
rzadp, line 5: We don't version docs

rzadp (approved):

PR #946, opened 2022-08-24
🦊 Add more metamask playwright tests

yivlad:


.github/workflows/CI.yml
```diff
@@ -55,12 +50,19 @@ jobs:
...
           set -e
+          echo "NEXT_PUBLIC_LOCALHOST_URL=http://localhost:8545" > .env.local
+          echo "NEXT_PUBLIC_MAINNET_URL=${MAINNET_URL}" >> .env.local
           pnpm dev &
           xvfb-run --auto-servernum --server-args="-screen 0 1280x960x24" -- pnpm test:playwright
         env:
           MAINNET_URL: ${{ secrets.MAINNET_PROVIDER_URL }}
           DEBUG: usedapp*
         working-directory: packages/example-next
+      - name: Cleanup
```
rzadp, line 61: We use github-hosted runners so no need for cleanup, but we can leave it in case we switch to self-hosted.

packages/core/src/hooks/useBlockMeta.ts
```diff
@@ -44,8 +44,16 @@ export function useBlockMeta(queryParams: QueryParams = {}) {
       }
   )
 
+  let timestamp: Date | undefined
+  try {
+    timestamp =
```
rzadp, line 49: Shouldn't we put this in `useMemo`? Won't it trigger re-renders if someone does:

```
const { timestamp } = useBlockMeta()

useEffect(xxx, [timestamp])
```
?

packages/core/src/hooks/useBlockMeta.ts
```diff
@@ -44,8 +44,16 @@ export function useBlockMeta(queryParams: QueryParams = {}) {
...
+      timestampResult !== undefined ? new Date(BigNumber.from(timestampResult.value).mul(1000).toNumber()) : undefined
+  } catch (e: any) {
+    console.warn('Failed to parse timestamp', e)
```
rzadp, line 52:
```suggestion
    console.warn('Failed to parse timestamp of a block.', e)
```

packages/example-next/.mocharc.json
```diff
@@ -0,0 +1,8 @@
+{
+  "require": "@swc-node/register",
+  "watchExtensions": "ts",
+  "extension": "ts",
+  "timeout": 60000,
+  "file": "./playwright/test-setup.ts",
+  "exit": true
```
rzadp, line 7: Is `exit: true` needed? See https://github.com/TrueFiEng/useDApp/pull/938

packages/core/src/hooks/useBlockMeta.ts
yivlad: Good catch

packages/core/src/hooks/useBlockMeta.ts
```diff
@@ -44,8 +45,19 @@ export function useBlockMeta(queryParams: QueryParams = {}) {
       }
   )
 
+  const timestamp = useMemo(() => {
+    let timestamp: Date | undefined
```
rzadp, line 49: Not needed, we can return from `try` and from `catch`

PR #953, opened 2022-09-01
⛓ Allow to use multichain with SIWE

jakvbs:


packages/siwe/src/provider.tsx
```diff
@@ -36,11 +36,10 @@ export interface SignInOptions {
 }
 
 export interface SiweProviderProps {
-  backendUrl?: string
```
rzadp, line 39: Why are you changing this? Is it necessary?

rzadp (approved):

PR #970, opened 2022-09-14
✂️ Disallow to use not configured chains in docs

jakvbs:


rzadp:
Done in https://github.com/TrueFiEng/useDApp/pull/971

PR #983, opened 2022-09-23
update Goerli chainId to 420

immunity20:
Hello Goerli Proof of Stake has  chainId equal to `420`


rzadp:
But the change has not happened **yet**?
It is still `5` for the time being, from what I can see.

immunity20:
Indeed sorry my fault :/

PR #988, opened 2022-09-27
🫘 Return error from useCall with invalid arguments

jakvbs:
Closes https://github.com/TrueFiEng/useDApp/issues/987


packages/core/src/hooks/useCall.test.tsx
```diff
@@ -420,6 +421,71 @@ describe('useCall', () => {
...
+                args: [123],
+              }),
+            {
+              config,
+            }
+          )
+          await waitForCurrent((val) => val !== undefined)
+
+          expect(result.current?.value).to.be.undefined
+          expect(result.current?.error).to.exist
```
rzadp, line 450: Check the exact error string

packages/core/src/helpers/calls.ts
```diff
@@ -17,6 +17,17 @@ export function warnOnInvalidCall(call: Call | Falsy) {
   console.warn(`Invalid contract call: address=${contract.address} method=${method} args=${args}`)
 }
 
+function getInvalidRawCall(call: Call, chainId: number): RawCall {
+  const { contract, method, args } = call
+
+  return {
+    address: contract.address,
+    data: `Invalid contract call: address=${contract.address} method=${method} args=${args}`,
```
rzadp, line 25: Consider doing `JSON.stringify(args)`

packages/core/src/helpers/calls.ts
```diff
@@ -20,14 +20,41 @@ export function warnOnInvalidCall(call: Call | Falsy) {
 /**
  * @internal Intended for internal use - use it on your own risk
  */
-export function encodeCallData(call: Call | Falsy, chainId: number, queryParams: QueryParams = {}): RawCall | Falsy {
+export interface InvalidRawCall {
+  chainId: ChainId
+  address: string
+  data: string
+  isValid: boolean
```
rzadp, line 27: This field does not make sense - there is no `InvalidRawCall` with `isValid=true`

rzadp: How about - `InvalidRawCall extends RawCall`
with an additional field `errorString`

packages/core/src/hooks/useRawCalls.ts
```diff
@@ -15,11 +16,11 @@ import { utils } from 'ethers'
  * @param calls List of calls, also see {@link RawCall}. Calls need to be in the same order across component renders.
  * @returns list of multicall calls. See {@link RawCallResult} and {@link useRawCall}.
  */
-export function useRawCalls(calls: (RawCall | Falsy)[]): RawCallResult[] {
+export function useRawCalls(calls: (RawCall | InvalidRawCall | Falsy)[]): RawCallResult[] {
   const { dispatchCalls, chains } = useContext(MultiChainStatesContext)
 
   useEffect(() => {
-    const filteredCalls = calls.filter(Boolean) as RawCall[]
+    const filteredCalls = calls.filter((call) => !!call && (call as InvalidRawCall).isValid !== false) as RawCall[]
```
rzadp, line 23:
```suggestion
    const filteredCalls = calls.filter((call) => !!call && !(errorString in call)) as RawCall[]
```

packages/core/src/helpers/calls.ts
```diff
@@ -20,29 +20,37 @@ export function warnOnInvalidCall(call: Call | Falsy) {
 /**
  * @internal Intended for internal use - use it on your own risk
  */
-export function encodeCallData(call: Call | Falsy, chainId: number, queryParams: QueryParams = {}): RawCall | Falsy {
+export function validateCall(call: Call | Falsy) {
   if (!call) {
-    return undefined
+    throw new Error('Call is falsy')
```
rzadp, line 25: Make sure the behaviour is unchanged.

Add a test for something like:
```
const result = useCall(undefined)
expect(result).to.xxx
```

And make sure the behaviour is the same before and after your changes.

packages/core/src/helpers/calls.ts
```diff
@@ -20,29 +20,37 @@ export function warnOnInvalidCall(call: Call | Falsy) {
...
+    throw new Error('Missing contract address or method name')
   }
+
   try {
-    const isStatic = queryParams.isStatic ?? queryParams.refresh === 'never'
-    const refreshPerBlocks = typeof queryParams.refresh === 'number' ? queryParams.refresh : undefined
+    contract.interface.encodeFunctionData(method, args)
+    return call
+  } catch (err: any) {
+    throw new Error(`Invalid contract call for method="${method}" on contract="${contract.address}": ${err.message}`)
```
rzadp, line 36: Try to include contract name in there, if it exists

jakvbs: I think it's not possible to extract name from Contract instance

rzadp (approved):

PR #991, opened 2022-09-28
🥃 Add checking calls order test in useCalls hook

jakvbs:


packages/core/src/hooks/useCall.test.tsx
```diff
@@ -420,6 +420,54 @@ describe('useCall', () => {
...
+              method: 'double',
+              args: [num],
+            }
+            const invalidCall = {
+              contract: doublerContract,
+              method: 'double',
+              args: ['invalid'],
+            }
+
+            return useCalls([validCall, null, validCall, invalidCall])
```
rzadp, line 441:
```suggestion
            return useCalls([validCall, null, invalidCall, validCall])
```

PR #997, opened 2022-10-06
🐙 Add e2e test for adding new network to Metamask

jakvbs:


yivlad (approved):

packages/docs/playwright/with-metamask.test.ts
```diff
@@ -85,6 +86,25 @@ describe(`Browser: ${browserType.name()} with Metamask`, () => {
...
+      const popupPromise = waitForPopup(context)
+      await page.click(XPath.text('button', 'Switch to Optimism'))
+      const popupPage = await popupPromise
+      await popupPage.click(XPath.text('a', 'View all'))
+
+      expect(await popupPage.isVisible(`//*[text()='${Optimism.chainName}']`)).to.be.true
+      expect(await popupPage.isVisible(`//*[text()='${Optimism.rpcUrl}']`)).to.be.true
+      expect(await popupPage.isVisible(`//*[text()='${Optimism.chainId}']`)).to.be.true
+      expect(await popupPage.isVisible(`//*[text()='${Optimism.nativeCurrency.symbol}']`)).to.be.true
+      expect(await popupPage.isVisible(`//*[text()='${Optimism.blockExplorerUrl}']`)).to.be.true
```
rzadp, line 106: How about adding - expect that "the data from this chain does not match metamask records" is not visible - I don't remember the message exactly, but you can recreate it by temporarily changing e.g. currency symbol

PR #1012, opened 2022-10-25
🥝 Add doc page about Tenderly forks

jakvbs:


packages/docs/docs/02-Guides/01-Connecting/09-Tenderly fork.mdx
```diff
@@ -0,0 +1,46 @@
+# Tenderly fork
+
+In some cases, you may want to connect to a network using a different chainId than its default. For example, your application is supposed to support the Mainnet network and forked Mainnet chain using [Tenderly](https://docs.tenderly.co/simulations-and-forks/simulation-api/testing/local-tests-on-top-of-mainnet-data-in-the-cloud/first-usage-of-tenderly-apis-in-tests), so the forked Mainnet should be accessible by a different number than the default chainId of the Mainnet, but the default for both networks it is the same which causes some problems.
```
rzadp, line 3:
```suggestion
In some cases, you may want to connect to a network using a different `chainId` than its default. For example, your application is supposed to support the Mainnet network and forked Mainnet chain using [Tenderly](https://docs.tenderly.co/simulations-and-forks/simulation-api/testing/local-tests-on-top-of-mainnet-data-in-the-cloud/first-usage-of-tenderly-apis-in-tests), so the forked Mainnet should be accessible by a different number than the default `chainId` of the Mainnet, but the default for both networks it is the same which causes some problems.
```

packages/docs/docs/02-Guides/01-Connecting/09-Tenderly fork.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+```tsx
+const config: Config = {
+  readOnlyChainId: Mainnet.chainId,
+  readOnlyUrls: {
+    [Mainnet.chainId]: 'https://mainnet.infura.io/v3/[your-infura-key]',
+    1337: 'https://rpc.tenderly.co/fork/[your-mainnet-fork-id]',
+  },
+  multicallAddresses: {
+    1337: Mainnet.multicall2Address ?? '0x5ba1e12693dc8f9c48aad8770482f4739beed696',
```
rzadp, line 17: Why `??`? It's either defined in our code or not?

packages/docs/docs/02-Guides/01-Connecting/09-Tenderly fork.mdx
```diff
@@ -0,0 +1,46 @@
...
+  },
+  multicallVersion: 2 as const,
+}
+```
+
+In this case, we want to have our forked Mainnet chain available on chainId: `1337`, which is the default for the Ganache network. But this won't work because it generates the provider with the same chainId as the default one, which is `1` for Mainnet.
```
rzadp, line 23:
```suggestion
In this case, we want to have our forked Mainnet chain available on `chainId`: `1337`, which is the default for the Ganache network. But this won't work because it generates the provider with the same `chainId` as the default one, which is `1` for Mainnet.
```

packages/docs/docs/02-Guides/01-Connecting/09-Tenderly fork.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+## Solution
+
+We can create a custom provider with overridden chainId:
```
rzadp, line 27:
```suggestion
We can create a custom provider with overridden `chainId`:
```

rzadp (approved):
Looks good, minor things to change.

packages/docs/docs/02-Guides/01-Connecting/09-Tenderly fork.mdx
```diff
@@ -0,0 +1,46 @@
+# Tenderly fork
```
rzadp, line 1: Consider naming it `Mainnet forking`.
The idea is to fork Mainnet, and tenderly is one of the tools to achievie it

packages/docs/docs/02-Guides/01-Connecting/09-Tenderly fork.mdx
jakvbs: It's because `multicall2Address` is an optional field in `Chain` model and its type is incompatible with type `string`, but in this case `Mainnet.multicall2Address as string` would be a better option.

PR #1014, opened 2022-10-31
Activate browser wallet check options type in runtime

yivlad:


.changeset/tidy-turtles-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+'@usedapp/core': patch
+---
+
+activateBrowserWallet check options type in runtime
```
rzadp, line 5: A little more descriptive for the user?

```suggestion
Safely handle invalid runtime parameters in `activateBrowserWallet`
```

rzadp (approved):

PR #1023, opened 2022-11-10
🥐 Keep transactionName in notifications after the transaction is mined

jakvbs:
Closes https://github.com/TrueFiEng/useDApp/issues/1022


rzadp (approved):
Add changeset

PR #1030, opened 2022-11-17
🍘 Update web3modal to V2

jakvbs:


derekbar90:
Excited for this

packages/example/webpack.config.js
```diff
@@ -27,14 +27,17 @@ module.exports = {
     new DefinePlugin({
       'process.env.MAINNET_URL': process.env.MAINNET_URL ? JSON.stringify(process.env.MAINNET_URL) : undefined,
       'process.env.LOCALHOST_URL': JSON.stringify(process.env.LOCALHOST_URL)
-    })
+    }),
+    new EnvironmentPlugin({
+      ...process.env,
```
rzadp, line 32: We should not do it that way, to make sure we're not leaking anything - change before potential merge.

Issue #1035, opened 2022-11-29
useCall with changing args causes global re-render of other useCall hooks


sirpy:
**Describe the bug**
I've noticed our component was triggering a re-render on a parent which shouldnt normally happen.
I've narrowed it down to the usage of `useCall` where `args` are changing based on input.

**To Reproduce**
I've created a reproducible sandbox. here: https://codesandbox.io/s/usedapp-usecall-refresh-bug-d9h7w9
press on the "call:" text (yes its clickable) this increases a counter, based on this counter an address is selected for the `balanceOf` call. This is the ContractCall component which is a child of Home.
You can see in the console that "Home" is printed every time `call:` is pressed indicating the parent Home component is re-rendering. This is because Home also have a `useCall`.
If you either remove useCall from Home or change in Contract.tsx `args: [addrs[x % 3]]` to `args: [addrs[0]]` you will see that pressing `call:` no longer re-renders Home.

**Software versions**
 - `useDapp` 1.22

sirpy:
looking at the useRawCalls hook, this seems to be because it is dependent on "chains" and this will change for every new useCall or new value returned. even though useMemo is used on the map extractCallResult, it will not memoize it since the result is a NEW array of possibly NEW result objects, and useMemo does not do deep equality, so even if the new array contains the same values, since it is a new array it will still be considered as different.
a possible solution is to use a more advanced memo hook like https://usehooks.com/useMemoCompar

sirpy:
@jakvbs @rzadp ?

sirpy:
#1045

PR #1045, opened 2022-12-14
✍️ Add usedeepmemo usecallback imrpovements

sirpy:
see #1035 for context
Added usage of deep equal in useRawCalls and useLogs which should reduce re-renders dependent on the results as input params.
Also improved useEthers not to return inline function but rather useCallback


sirpy:
#1035 
@rzadp

rzadp:
@pawelpolak2 You could contribute by adding a clever test that would count the number of rerenders.
The test should be red before this change and green after this change.

sirpy:
@rzadp sorry don't have the bandwidth for that. i've verified the deep compare works so things like logs do not change, and that it improved our re-renders also the usecallback improvements are obvious but i've also verified they stopped some of the re-renders we were having

rzadp:
> @rzadp sorry don't have the bandwidth for that.

Yes that is completely fine, please note that I have tagged someone else for that :)

yivlad (approved):

sirpy:
@yivlad why was this closed?

yivlad:
@sirpy most of the changes were already added in other PRs. The only thing that is not implemented is `useDeepMemo` hook, and I played around with it a bit and it doesn't really seem to help. Besides I would prefer not to add additional external dependencies.

sirpy:
@yivlad 
first can you point me to the PR where useEtherrs is fixed to use useCallback correctly + deps?

secondly i'm not sure what you mean by it doesnt really seem to help, you are probably not testing things correctly.
you do understand that useMemo has absolutely 0 effect when dealing with arrays and objects that are being re-created on each render?

```
const logs = useLogs()
<SomeComponent logs={logs}/>
```
in the above case without the deep memo SomeComponent would be re-rendered everytime your global store changes.

the only case where usedeepmemo would in did not have much effect is inside useRawCalls, because that depends on "chains" directly.

yivlad:
@sirpy 
Here is [the PR](https://github.com/TrueFiEng/useDApp/pull/1058) I'm talking about.

Giving this a second thought though I think it actually would make sense to use `useDeepMemo` in the `useLogs` hook. I think this is also what you're saying as well? If so, you can  open another PR with this changes? I would also like to use [this](https://github.com/apollographql/apollo-client/blob/master/src/react/hooks/utils/useDeepMemo.ts) implementation of the useDeepMemo hook. It introduces less external dependencies and also handles the case of cycle references.

sirpy:
@yivlad well it makes sense to use it anywhere you use "usememo" with non primitives like arrays/objects.
I dont see how this is less dependencies, unless you mean you want to include the source code.

yivlad:
@sirpy exactly, I think the best option is to copy the code of the `useDeepMemo` hook. We'll have to introduce only one dependency that doesn't have any other dependencies (`@wry/equality`)

PR #1046, opened 2022-12-14
✍️ Add pending sig notification

sirpy:
Add a new notificaiton type `transactionPendingSignature`
added the transactionRequest object to promiseTransaction and to the notification so the developer can have the full context as the transaction itself is not available yet


sirpy:
@rzadp 
looking at line 80 of usePromiseTransaction
```
  useEffect(() => {
    return () => {
      gnosisSafeContract?.removeAllListeners()
    }
  }, [gnosisSafeContract])
```

I'm not sure this will work as expected as gnosisSafeContract is assigned inside a usecallback and not via setState

yivlad (approved):

PR #1047, opened 2022-12-14
✍️ Add useSigner hook

### LINKED ISSUE #1032: Fallback does not have send() method
**Describe the bug**
I upgrade from the version 1.1.5 to the newer one 1.2.2 and in the parts of the code that I used `Web3Ethers` I'm unable to keep using due to an upgrade in the code of TrueFiEng. ( which I needed to upgrade to get the newer arbitrum chain goerli )

**To Reproduce**
make a send with the library in the ` useWeb3()` function: 
image

**Software versions**
 - `useDapp` version (installed version of `@usedapp/core`): 1.2.2
 - Package manager : yarn 3,1,9
 - Node version (`node --version`): v16.14.2

**Additional context**

in the previous version we didn't had fallback provider in useWeb3():
image
Previous:
image
Current:
image

Fallback does not have send function. I'm not sure how to workaround or deal with it yet. Is there a way or suggestion you guys could make ? ( What I planned on doing was forking and removing fallback )

pawelpolak2:
Closes #1032


packages/core/src/hooks/useSigner.tsx
```diff
@@ -0,0 +1,15 @@
+import { useEthers } from './useEthers'
+import { useState, useEffect } from 'react'
+import { JsonRpcProvider, JsonRpcSigner } from '@ethersproject/providers'
+
+export function useSigner(): JsonRpcSigner | undefined {
+  const { library, account } = useEthers()
+  const [signer, setSigner] = useState<JsonRpcSigner | undefined>()
+
+  useEffect(() => {
+    if (account) setSigner((library as JsonRpcProvider).getSigner())
```
rzadp, line 10: It might be safer to do:

```
if (library !== undefined && 'getSigner' in library)
```

Because if for some reason `account` would update faster than `library` you would have a problem.
And this way TS is smart enough and you don't have to use `as`

packages/core/src/hooks/useSigner.tsx
```diff
@@ -0,0 +1,15 @@
+import { useEthers } from './useEthers'
+import { useState, useEffect } from 'react'
+import { JsonRpcProvider, JsonRpcSigner } from '@ethersproject/providers'
+
+export function useSigner(): JsonRpcSigner | undefined {
```
rzadp, line 5: If you want to release it to the public, you need:
- a changeset
- Add JSDOC
    - With some human readable comments that will end up in docs - check that they are properly auto-generated in the API reference section
    - with a `@public` annotation
- Export it in https://github.com/TrueFiEng/useDApp/blob/master/packages/core/src/index.ts
    - We don't export `*` wildcards in order to consciously control public API contract

rzadp:
@pawelpolak2 You can change the description to close this: https://github.com/TrueFiEng/useDApp/issues/1032 using [keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

After this gets merged & released, you can post in this issue with a link to the docs of the hook and a number of new version that has it.

PR #1053, opened 2023-01-04
➕ Add Rave Names resolution

Pokesi:
Add [Rave Names](https://rave.domains/) resolution into the `useLookupAddress` hook.

The hook will by default resolve to ENS, but if a second optional parameter is set to `true`, the hook will resolve to a Rave Name.

### DIFF
--- a/packages/core/src/hooks/useLookupAddress.ts
+++ b/packages/core/src/hooks/useLookupAddress.ts
@@ -1,11 +1,13 @@
 import { useEffect, useState } from 'react'
 import { useEthers } from './useEthers'
+import { Contract } from 'ethers'
 
 /**
- * `useLookupAddress` is a hook that is used to retrieve the ENS (e.g. `name.eth`) for a specific address.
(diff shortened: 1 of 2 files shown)


packages/core/src/hooks/useLookupAddress.ts
```diff
@@ -17,22 +19,30 @@ import { useEthers } from './useEthers'
...
 
     void (async () => {
       if (!library || !address) return
       try {
-        setIsLoading(true)
-        const resolved = await library.lookupAddress(address)
-        if (!mounted) return
-        setENS(resolved)
+	setIsLoading(true)
+	if (rave) {
```
yivlad, line 36: I don't think this will work as intended as `rave` variable here is of type `Contract`, not `boolean`. Please move declaration of the rave contract from line 27 inside this `if` clause.

Pokesi:
Updated: https://github.com/TrueFiEng/useDApp/pull/1053/commits/f59a0b1eef14d2f59054f7a0d6b82f7c356ab574

yivlad (approved):

PR #1058, opened 2023-01-10
♠️ Move useEthers to context

yivlad:


packages/core/src/providers/network/connectors/context.tsx
```diff
@@ -32,6 +69,13 @@ export const ConnectorContext = createContext<ConnectorContextValue>({
   //eslint-disable-next-line @typescript-eslint/no-empty-function
   reportError: () => {},
   isLoading: false,
+  setError: () => {
+    throw new Error('Function not implemented.')
+  },
+  active: false,
+  switchNetwork: () => {
+    throw new Error('Function not implemented.')
+  },
```
sirpy, line 78: should use usecallback
@yivlad

packages/core/src/providers/network/connectors/context.tsx
```diff
@@ -117,23 +196,88 @@ export function ConnectorContextProvider({ children }: ConnectorContextProviderP
...
+        setProvider(controller.getProvider())
+        if (accounts[0]) {
+          setAccount(getAddress(accounts[0]))
+        } else {
+          setAccount(undefined)
+        }
+      }
+      setErrors([...errors])
+    })
+  }, [controller, controller?.getProvider()])
```
sirpy, line 233: @yivlad missing deps for the set* functions

packages/core/src/providers/network/connectors/context.tsx
```diff
@@ -117,23 +196,88 @@ export function ConnectorContextProvider({ children }: ConnectorContextProviderP
...
+
+    for (const networkState of Object.values(networkStates)) {
+      if (networkState.errors.length > 0) {
+        setError(networkState.errors[networkState.errors.length - 1])
+        return
+      }
+    }
+
+    setError(errors?.[errors.length - 1])
+  }, [chainId, errors, networkStates])
```
sirpy, line 262: missing deps for seterror

packages/core/src/providers/network/connectors/context.tsx
yivlad: Not necessary

packages/core/src/providers/network/connectors/context.tsx
yivlad: Not necessary

packages/core/src/providers/network/connectors/context.tsx
yivlad: It should not, this value that the context is initiallized with

PR #1060, opened 2023-01-11
📦 Update @metamask/detect-provider to 2.0.0

gsteenkamp89:
# Bump @metamask/detect-provider version to 2.0.0

This will resolve [this](https://github.com/TrueFiEng/useDApp/issues/529) issue relating to sourcemaps.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -17,7 +17,7 @@
     "usedapp-generate-hooks": "./bin/usedapp-generate-hooks.js"
   },
   "dependencies": {
-    "@metamask/detect-provider": "2.0.0",
+    "@metamask/detect-provider": "^2.0.0",
     "@uniswap/token-lists": "^1.0.0-beta.27",
     "fetch-mock": "^9.11.0",
     "lodash.merge": "^4.6.2",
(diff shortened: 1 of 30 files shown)


packages/core/package.json
```diff
@@ -17,7 +17,7 @@
     "usedapp-generate-hooks": "./bin/usedapp-generate-hooks.js"
   },
   "dependencies": {
-    "@metamask/detect-provider": "^1.2.0",
+    "@metamask/detect-provider": "2.0.0",
```
yivlad, line 20: Is there a reason why the version is fixed?

yivlad (changes requested):
Please also add changeset

packages/core/package.json
gsteenkamp89: shouldn't be, I'll add the caret back! 👍

yivlad (commented):
@gsteenkamp89 please also update the lockfile - CI seems to fail because of this

gsteenkamp89:
> @gsteenkamp89 please also update the lockfile - CI seems to fail because of this

@yivlad done 👍

yivlad (approved):

tamebadger:
what's the timeline on a release that includes this fix out of interest sake @yivlad ?

yivlad:
Hey @tamebadger, we'll publish the latest version to npm when we'll be sure that the version is stable. For now you're always welcome to download `dev` version from npm - it's the latest master from this repo.

PR #1110, opened 2023-04-27
🔵 Adding Base Goerli Testnet to chains id

afa7789:
Guys I need it please.

I got all the data needed for it in: https://docs.base.org/


afa7789:
I'm missing the Multicall contract, I will deploy one and ping here.

packages/core/src/model/chain/index.ts
```diff
@@ -29,3 +29,4 @@ export * from './arbitrumReddit'
 export * from './rootstock'
 export * from './klaytn'
 export * from './celo'
+export * from './base'
```
nezouse, line 32:
```suggestion
export * from './base'

```

nezouse (approved):

PR #1111, opened 2023-05-02
📜 Add scroll alpha testnet

afa7789:
All the data needed for this PR is shown here: https://guide.scroll.io/user-guide/setup.

I really need you guys to at least review my PR's there are at list more 5 chains in which I will need to add their ID to useDApp.

Please merge, my previous one, this one is considering that I have one chain more ( as If we merged base goerli ), reducing the stress of merge conflicts. ( which will happen )


packages/core/src/model/index.ts
```diff
@@ -4,4 +4,4 @@ export * from './Currency'
 export * from './CurrencyValue'
 export * from './TransactionStatus'
 export * from './TransactionOptions'
-export * from './types'
+export * from './types'
```
nezouse, line 7:
```suggestion
export * from './types'

```

nezouse (approved):

PR #1142, opened 2023-09-12
🐳 Add kroma network

afa7789:
https://kroma.network/ all the data was captured from then.

it's relatively new chain, not even have verified contracts at kroma network blockscout.

### DIFF
--- a/.changeset/adding_kroma.md
+++ b/.changeset/adding_kroma.md
@@ -1,3 +1,7 @@
+---
+"@usedapp/core": patch
+---
+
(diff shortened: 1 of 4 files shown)


.changeset/adding_kroma.md
```diff
@@ -0,0 +1,7 @@
+Adding the network kroma mainnet:
```
nezouse, line 1:
```suggestion
---
"@usedapp/core": patch
---

Adding the network kroma mainnet:
```

nezouse (approved):

PR #1151, opened 2024-04-21
🔬 Make useRawLogs more efficient and responsive

TylerEther:
- Makes use of isStatic query param
- Makes use of refresh query param and global config
- Clears logs when there's a change in the filter to minimize the amount of time invalid old logs are returned
- Prevents some race conditions
- It no longer fetches logs with every re-render


Szymx95 (commented):
Hey @TylerEther,

thank you for contributing to the repo.

I have a question regarding the PR

Removing Promise from useRawLogs does simplify the code however introduces breaking changes so we would have to release this PR as new major version. Is there a way you could keep the function parameters unchanged while introducing the rest of your changes ?

packages/core/src/hooks/useRawLogs.ts
```diff
@@ -25,21 +26,82 @@ export function useRawLogs(
...
+      setLastTopics(!filter ? undefined : JSON.stringify(filter.topics))
+      setLastChainId(chainId)
+      setLastBlockNumber(blockNumber)
+    } finally {
+      setIsLoading(false)
+    }
   }
 
-  useEffect(() => {
+  useMemo(() => {
```
Szymx95, line 102: Also shouldn't `useEffect` behave the same as `useMemo` here. Why the change ?

TylerEther: Good catch. I was playing around with what worked best. I've switched back to `useEffect` as they behave the same and to also handle cleanup procedures.

TylerEther:
> Hey @TylerEther,
> 
> thank you for contributing to the repo.
> 
> I have a question regarding the PR
> 
> Removing Promise from useRawLogs does simplify the code however introduces breaking changes so we would have to release this PR as new major version. Is there a way you could keep the function parameters unchanged while introducing the rest of your changes ?

Good point. It was tricky, but I found a way to handle filter promises so I added that back.

TylerEther:
I've introduced a helper hook, `useResolvedFilter` and helper functions, `deepEqual` and `isPrimitive`. Let me know if you'd like me to extract these into a common helper file or if you'd like some unit tests for them.

packages/core/src/hooks/useRawLogs.ts
```diff
@@ -1,10 +1,75 @@
...
 import type { Filter, FilterByBlockHash, Log } from '@ethersproject/abstract-provider'
 import { Falsy } from '../model/types'
+import { ChainId } from '../constants'
+
+function deepEqual(obj1: any, obj2: any) {
+  if (obj1 === obj2)
+    // it's just the same object. No need to compare.
+    return true
+
+  if (obj1 == null) return obj1 == null
```
Szymx95, line 15:
```suggestion
  if (obj1 == null) return obj2 == null
```
I believe this is a typo

otherwise `deepEqual(null, 2)` would return true

packages/core/src/hooks/useRawLogs.ts
```diff
@@ -1,10 +1,75 @@
...
+import { useBlockNumbers, useBlockNumber, useConfig } from '../hooks'
 import { QueryParams } from '../constants/type/QueryParams'
 import type { Filter, FilterByBlockHash, Log } from '@ethersproject/abstract-provider'
 import { Falsy } from '../model/types'
+import { ChainId } from '../constants'
+
+function deepEqual(obj1: any, obj2: any) {
+  if (obj1 === obj2)
+    // it's just the same object. No need to compare.
+    return true
```
Szymx95, line 13:
```suggestion
  if (obj1 === obj2) return true
```
I believe it's self explanatory enough no need to add comment

packages/core/src/hooks/useRawLogs.ts
Szymx95: I don't know if this is intended behaviour  but it seems it would return 
`deepEqual(null, undefined)` as true

packages/core/src/hooks/useRawLogs.ts
```diff
@@ -1,10 +1,75 @@
...
+    let active = true // Flag to prevent setting state after unmount
+
+    const resolveFilter = async () => {
+      let _filter: Filter | FilterByBlockHash | Falsy = undefined
+
+      if (filter instanceof Promise) {
+        _filter = await filter
+      } else {
+        _filter = filter
+      }
```
Szymx95, line 55:
```suggestion
      let _filter: Filter | FilterByBlockHash | Falsy = await filter
```
If you await non promise value it will just return it as value otherwise it will resolve promise

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#return_value

Szymx95 (commented):
Thank you for taking your time and reworking the solution.

I've found a few things in `deepEqual` that I think might lead to unintended behaviours. It would be great if you could extract those helpers and write some unit tests.

packages/core/src/hooks/useRawLogs.ts
TylerEther: Good catch!

packages/core/src/hooks/useRawLogs.ts
TylerEther: Clean

TylerEther:
Thanks for the review @Szymx95. I've added tests for everything, made some improvements, and additionally tested the latest changes with my web app. Everything functions well.

Szymx95:
Looks good could you rebase to current master branch ?

TylerEther:
> Looks good could you rebase to current master branch ?

Done @Szymx95

JustynaBroniszewska (approved):

PR #1152, opened 2024-04-23
Update the gnosis safe api url.

robotoer:
The address used to access the gnosis safe API changed in 2023: https://forum.safe.global/t/announcement-to-all-builders-using-safe-services/3225

### DIFF
--- a/packages/core/src/helpers/gnosisSafeUtils.ts
+++ b/packages/core/src/helpers/gnosisSafeUtils.ts
@@ -88,9 +88,9 @@ export const calculateSafeTransactionHash = (
 export const getLatestNonce = async (chainId: number, safeAddress: string): Promise<number | null | undefined> => {
   try {
     const response = await fetch(
-      `https://safe-transaction.${
+      `https://safe-transaction-${
         getChainById(chainId)?.chainName
-      }.gnosis.io/api/v1/safes/${safeAddress}/all-transactions?limit=1&executed=false&queued=true`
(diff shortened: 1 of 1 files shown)


robotoer:
For those looking for an immediate solution, try using https://www.npmjs.com/package/patch-package with the following patch file (`patches/@usedapp+core+1.2.13.patch`):

```patch
diff --git a/node_modules/@usedapp/core/src/helpers/gnosisSafeUtils.ts b/node_modules/@usedapp/core/src/helpers/gnosisSafeUtils.ts
index 3225279..c4db3b6 100644
--- a/node_modules/@usedapp/core/src/helpers/gnosisSafeUtils.ts
+++ b/node_modules/@usedapp/core/src/helpers/gnosisSafeUtils.ts
@@ -88,9 +88,9 @@ export const calculateSafeTransactionHash = (
 export const getLatestNonce = async (chainId: number, safeAddress: string): Promise<number | null | undefined> => {
   try {
     const response = await fetch(
-      `https://safe-transaction.${
+      `https://safe-transaction-${
         getChainById(chainId)?.chainName
-      }.gnosis.io/api/v1/safes/${safeAddress}/all-transactions?limit=1&executed=false&queued=true`
+      }.safe.global/api/v1/safes/${safeAddress}/all-transactions?limit=1&executed=false&queued=true`
     )
     if (!response.ok) return null
     const allTransactions = await response.json()
```

nezouse (approved):

PR #1157, opened 2024-05-13
🧚🏻 Remove goerli from example app  multichain section

nezouse:


packages/example/src/entrypoint.tsx
```diff
@@ -18,6 +19,7 @@ import { getDefaultProvider } from '@ethersproject/providers'
 
 const readOnlyUrls: Config['readOnlyUrls'] = {
   [Mainnet.chainId]: process.env.MAINNET_URL || getDefaultProvider('mainnet'),
+  [Sepolia.chainId]: 'https://rpc.sepolia.org',
   [Goerli.chainId]: process.env.MAINNET_URL
```
JustynaBroniszewska, line 23: Shouldn't we remove it?

PR #1173, opened 2025-02-05
🔵 Update Base RPC url

### LINKED ISSUE #1172: Base Network RPC is Wrong
**Describe the bug**
The Primary Base RPC is `https://mainnet.base.org/`.
You can see this in their docs here: https://docs.base.org/docs/network-information/
However, currently the RPC for Base is set to `https://base.org` which is wrong.
You can see this here: https://github.com/TrueFiEng/useDApp/blob/e0c8437b0726ff2c6b1b4772cf031d46941530e8/packages/core/src/model/chain/base.ts#L31C12-L31C28

The impact of this is that when you try to add the Base network using useDapp, it errors in the wallet.

**To Reproduce**
Install Metamask, ensure you don't have the Base network already added.
Import the Switch Network function `const { switchNetwork } = useEthers()`.
Call `switchNetwork(8453)`.
You will get this error:

Image

**Software versions**
Happens with the latest version of useDapp `1.2.16`.
Happens with yarn, npm, or anything.

chase-manning:
Fixes: #1172 

TLDR: Base Mainnet RPC is wrong, this causes issues adding the Base network via the `switchNetwork` function. PR updates it to the correct RPC from their docs: https://docs.base.org/docs/network-information/

### DIFF
--- a/.changeset/heavy-fishes-whisper.md
+++ b/.changeset/heavy-fishes-whisper.md
@@ -1,0 +1,5 @@
+---
+"@usedapp/core": patch
+---
+
+Update Base RPC url
--- a/packages/core/src/model/chain/base.ts
+++ b/packages/core/src/model/chain/base.ts
@@ -28,7 +28,7 @@ export const Base: Chain = {
   isTestChain: false,
   isLocalChain: false,
   multicallAddress: '0x38641b7a50CDcfde75a7A83eB7c02581F3279362',
-  rpcUrl: 'https://base.org',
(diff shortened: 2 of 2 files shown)


nezouse (approved):

0xrobin2192:
Hi @nezouse - do you need anything from us in order to merge this? Thanks!

nezouse:
Can't really do much as tests are failing, i no longer have maintainer role on this repo to skip the tests and don't really have time to fix them

